PR #6, opened 2014-12-30
Use minimatch to handle negative globs, fixes #2; Run positive globs in parallel, fixes #3

### LINKED ISSUE #2: Negative globs handling is very slow
See this [benchmark](https://gist.github.com/UltCombo/27adde9fc0a6d511e909) comparing [glob-stream](https://github.com/wearefractal/glob-stream) with globby. Although glob-stream has the overhead of creating streams for the matched files (which is not really significant as there is only 1 matched file in my test), glob-stream is still over 2 times faster than globby.

This is because globby seeks the paths that would be matched by the negative globs generating too many unnecessary file system calls, while glob-stream uses minimatch to handle negative globs.

### LINKED ISSUE #3: Multiple positive globs handling is slow
globby currently uses [`async.reduce`](https://github.com/caolan/async#reduce) which runs in series, I believe it would be better to run in parallel.

UltCombo:
Fixes #2 
Fixes #3 

Negative globs handling is over 2.5x faster now in regards to my previous benchmark.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
 var union = require('array-union');
 var async = require('async');
 var glob = require('glob');
-var minimatch = require('minimatch');
+var Minimatch = require('minimatch').Minimatch;
 
 function arrayify(arr) {
 	return Array.isArray(arr) ? arr : [arr];
@@ -40,20 +40,20 @@ module.exports = function (patterns, opts, cb) {
 					return;
 				}
 
-				var negativePatterns = negatives.filter(function (negative) {
+				var negativeMatchers = negatives.filter(function (negative) {
 					return negative.index > positive.index;
 				}).map(function (negative) {
-					return negative.pattern;
+					return new Minimatch(negative.pattern, opts);
 				});
 
-				if (negativePatterns.length === 0) {
+				if (negativeMatchers.length === 0) {
 					callback(null, paths);
 					return;
 				}
 
 				callback(null, paths.filter(function (path) {
-					return negativePatterns.every(function (pattern) {
-						return minimatch(path, pattern, opts);
+					return negativeMatchers.every(function (matcher) {
+						return matcher.match(path);
 					});
 				}));
 			});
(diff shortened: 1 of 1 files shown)


UltCombo:
@sindresorhus 

> Wonder if there's more opportunities for perf improvements.

Oh, I just found one more small optimization we can make (see wearefractal/glob-stream#33), but it shouldn't matter much. I'll push it to globby later if it doesn't make the code too ugly.

I believe the next big optimization would be to use node-glob's [`ignore` API](https://github.com/isaacs/node-glob/issues/115), but that is not implemented yet. `;)`

As for benchmarking, I'll open an issue to discuss some possible improvements.

UltCombo:
https://github.com/sindresorhus/globby/commit/e402d9fbb6d16f2c4f1464d8323fefd16116af04

sindresorhus:
> Did you update the globby dependency after merging?

Yes, ofc. npm caches branches so you kinda have to do it.

> I believe the next big optimization would be to use node-glob's ignore API, but that is not implemented yet. ;)

That will help a lot! :)

-

Seeing how Isaac don't want to support multiple patterns natively it seems that this module will have to stay.

UltCombo:
Oh I'm also running into the same perf diff issue, looks like the first bench is up to 10% slower for some reason.

If I add another benchmark test before the actual benchmarks then it even outs the difference a bit:

```
                     negative globs (100 negated paths)
            227 op/s » warm up (ignore)
            240 op/s » globby async (working directory)
            244 op/s » globby async (upstream/master)
```

UltCombo:
I fiddled a bit with [benchmark.js](http://benchmarkjs.com/), and according to it, upstream/master is 10% slower hahah.
This benchmarking stuff is too much headache.

sindresorhus:
@UltCombo hmm, maybe it has something to do with the filesystem/harddrive cache.

UltCombo:
That's the most likely yeah, seeing as filesystem access is the biggest bottleneck here.

PR #11, opened 2015-01-01
Benchmark improvements

UltCombo:
How can we improve the benchmarking suite?

The main pain point I see at the moment is having to update the globby upstream/master and glob-stream dependencies manually. Those are currently pulled as [npm dependencies from GitHub](https://github.com/sindresorhus/globby/blob/master/package.json#L59-L60), but that's far from ideal.

I'm considering to add some `child_process.spawn`/`exec` before running the benchs to automate the update process. Having to clear those packages from the npm cache and download them entirely again doesn't seem efficient though, I'm considering pulling them from Git directly (`git clone` if the dependency is not present, `git pull` otherwise). If using Git, then `npm install` would have to ensued in those packages to download their dependencies, so this may be overcomplicating it. I don't feel like downloading those entire packages before every benchmark run though.

Ideas?

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "mocha && matcha bench.js"
+    "bench": "npm update globby glob-stream && matcha bench.js",
+    "test": "mocha"
   },
   "files": [
     "index.js"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
 language: node_js
 node_js:
   - '0.10'
+script: 'npm test && npm run bench'


UltCombo:
Transformed this issue into a PR.
@sindresorhus Please take a look/suggest improvements. `=]`

package.json
```diff
@@ -13,7 +13,8 @@
     "node": ">=0.10.0"
   },
   "scripts": {
-    "test": "mocha && matcha bench.js"
+    "bench": "npm up globby glob-stream && matcha bench.js",
```
sindresorhus: Can you use the full name `up` => `update`?

UltCombo: of course, done.

PR #17, opened 2015-08-26
Use a Promise API instead of a callback.

wbinnssmith:
As discussed in sindresorhus/del#29, here's a Promise implementation for globby. I followed the same direction as with that PR, removing the callback interface. I don't mind adding it back though.

Using Promises alone I was able to get rid of the `async` dependency and clean up the implementation quite a bit. ~~I did have to add a small wrapper around `glob` inline, but considered a [ponyfill-compatible, standalone version of denodeify](https://github.com/valeriangalliat/es6-denodeify). There's also [ahmadnassri/glob-promise](https://github.com/ahmadnassri/glob-promise), but it won't take an arbitrary Promise constructor and would force users to have a global one.~~ Using `pify`.

Let me know where to go from here. Totally happy to adapt.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,9 +1,10 @@
 'use strict';
+var Promise = require('pinkie-promise');
 var arrayUnion = require('array-union');
 var objectAssign = require('object-assign');
-var async = require('async');
 var glob = require('glob');
 var arrify = require('arrify');
+var pify = require('pify');
 
 function sortPatterns(patterns) {
 	patterns = arrify(patterns);
@@ -38,37 +39,19 @@ function setIgnore(opts, negatives, positiveIndex) {
 	return opts;
 }
 
-module.exports = function (patterns, opts, cb) {
+module.exports = function (patterns, opts) {
 	var sortedPatterns = sortPatterns(patterns);
-
-	if (typeof opts === 'function') {
-		cb = opts;
-		opts = {};
-	}
+	opts = opts || {};
 
 	if (sortedPatterns.positives.length === 0) {
-		cb(null, []);
-		return;
+		return Promise.resolve([]);
 	}
 
-	async.parallel(sortedPatterns.positives.map(function (positive) {
-		return function (cb2) {
-			glob(positive.pattern, setIgnore(opts, sortedPatterns.negatives, positive.index), function (err, paths) {
-				if (err) {
-					cb2(err);
-					return;
-				}
-
-				cb2(null, paths);
-			});
-		};
-	}), function (err, paths) {
-		if (err) {
-			cb(err);
-			return;
-		}
-
-		cb(null, arrayUnion.apply(null, paths));
+	return Promise.all(sortedPatterns.positives.map(function (positive) {
+		var globOpts = setIgnore(opts, sortedPatterns.negatives, positive.index);
+		return pify(glob, Promise)(positive.pattern, globOpts);
+	})).then(function (paths) {
+		return arrayUnion.apply(null, paths);
 	});
 };
 
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # globby [![Build Status](https://travis-ci.org/sindresorhus/globby.svg?branch=master)](https://travis-ci.org/sindresorhus/globby)
 
-> Extends [glob](https://github.com/isaacs/node-glob) with support for multiple patterns
+> Extends [glob](https://github.com/isaacs/node-glob) with support for multiple patterns and exposes a Promise API
 
 
 ## Install
@@ -30,7 +30,7 @@ globby(['*', '!cake'], function (err, paths) {
 
 ## API
 
-### globby(patterns, [options], callback)
+### globby(patterns, [options])
 
 ### globby.sync(patterns, [options])
 
@@ -47,8 +47,8 @@ Type: `object`
 
 See the node-glob [options](https://github.com/isaacs/node-glob#options).
 
-#### callback(err, paths)
-
+`globby.sync` synchronously returns an array of paths matching the glob expression.
+`globby` returns a Promise object that resolves to an array of matching paths or is rejected with an error.
 
 ## Globbing patterns
 
--- a/test.js
+++ b/test.js
@@ -22,27 +22,21 @@ after(function () {
 	fixture.forEach(fs.unlinkSync.bind(fs));
 });
 
-it('should glob - async', function (cb) {
-	globby('*.tmp', function (err, paths) {
-		assert(!err, err);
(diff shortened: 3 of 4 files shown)


index.js
```diff
@@ -1,10 +1,22 @@
...
 var objectAssign = require('object-assign');
-var async = require('async');
 var glob = require('glob');
 var arrify = require('arrify');
 
+function globP(pattern, opts) {
+	return new Promise(function (resolve, reject) {
+		glob(pattern, opts, function (err, files) {
+			if (err) {
+				reject(err);
```
UltCombo: Might be good to `return` after rejecting.
Although ES2015 promises' can't mutate their state after being settled (rejected or fulfilled), it looks odd to call both `reject` and `resolve` when there's an error.

wbinnssmith: Good catch. Fixed

index.js
```diff
@@ -38,37 +50,19 @@ function setIgnore(opts, negatives, positiveIndex) {
...
-		if (err) {
-			cb(err);
-			return;
-		}
-
-		cb(null, arrayUnion.apply(null, paths));
+	return Promise.all(sortedPatterns.positives.map(function (positive) {
+		return globP(positive.pattern, setIgnore(opts, sortedPatterns.negatives, positive.index));
+	})).then(function (paths) {
+		return arrayUnion.apply(null, paths);
```
sindresorhus: No need for a `return` here.

index.js
```diff
@@ -1,10 +1,22 @@
 'use strict';
+var Promise = require('pinkie-promise');
 var arrayUnion = require('array-union');
 var objectAssign = require('object-assign');
-var async = require('async');
 var glob = require('glob');
 var arrify = require('arrify');
 
+function globP(pattern, opts) {
```
sindresorhus: Can you just put this inline?

index.js
```diff
@@ -1,10 +1,22 @@
...
+	return new Promise(function (resolve, reject) {
+		glob(pattern, opts, function (err, files) {
+			if (err) {
+				return reject(err);
+			}
+
+			return resolve(files);
```
sindresorhus: No need for a `return` here.

index.js
```diff
@@ -1,10 +1,22 @@
...
-var async = require('async');
 var glob = require('glob');
 var arrify = require('arrify');
 
+function globP(pattern, opts) {
+	return new Promise(function (resolve, reject) {
+		glob(pattern, opts, function (err, files) {
+			if (err) {
+				return reject(err);
+			}
```
sindresorhus:
```js
if (err) {
  reject(err);
  return;
}

File: readme.md [hunk 6]
```diff
@@ -1,6 +1,6 @@
 # globby [![Build Status](https://travis-ci.org/sindresorhus/globby.svg?branch=master)](https://travis-ci.org/sindresorhus/globby)
 
-> Extends [glob](https://github.com/isaacs/node-glob) with support for multiple patterns
+> Extends [glob](https://github.com/isaacs/node-glob) with support for multiple patterns and exposes a Promise API.
```
sindresorhus: No dot at the end.

sindresorhus:
Looks good. I've added some nitpick inline comments.

index.js
wbinnssmith: Are you sure? Without the return the promise won't yield the list of deleted paths.
```
   1) should glob - async
   2) should glob with multiple patterns - async
   3) should respect patterns order - async
   ✓ should glob - sync
   ✓ cwd option
   ✓ should not mutate the options object - async
   ✓ should not mutate the options object - sync

  4 passing (29ms)
  3 failing

  1)  should glob - async:
     AssertionError: undefined deepEqual [ 'a.tmp', 'b.tmp', 'c.tmp', 'd.tmp', 'e.tmp' ]
      at test.js:27:10

  2)  should glob with multiple patterns - async:
     AssertionError: undefined deepEqual [ 'a.tmp', 'b.tmp' ]
      at test.js:33:10

  3)  should respect patterns order - async:
     AssertionError: undefined deepEqual [ 'a.tmp' ]
      at test.js:39:10
```

sindresorhus: Oh, never mind, I was reading it wrong.

wbinnssmith:
Awesome. Thanks for the feedback. Should be addressed now.

index.js
```diff
@@ -38,37 +38,29 @@ function setIgnore(opts, negatives, positiveIndex) {
...
 	var sortedPatterns = sortPatterns(patterns);
-
-	if (typeof opts === 'function') {
-		cb = opts;
-		opts = {};
-	}
+	opts = opts || {};
 
 	if (sortedPatterns.positives.length === 0) {
 		cb(null, []);
```
sindresorhus: You forgot to handle this case ;)

index.js
```diff
@@ -38,37 +38,29 @@ function setIgnore(opts, negatives, positiveIndex) {
...
 		return;
 	}
 
-	async.parallel(sortedPatterns.positives.map(function (positive) {
-		return function (cb2) {
-			glob(positive.pattern, setIgnore(opts, sortedPatterns.negatives, positive.index), function (err, paths) {
+	return Promise.all(sortedPatterns.positives.map(function (positive) {
+		return new Promise(function (resolve, reject) {
```
sindresorhus: Use https://github.com/sindresorhus/pify

```js
var opts = setIgnore(opts, sortedPatterns.negatives, positive.index);
return pify(glob, Promise)(positive.pattern, opts);
```

index.js
wbinnssmith: eep.

index.js
wbinnssmith: You literally just made this module to clean this up didn't you?

test.js
```diff
@@ -52,6 +46,16 @@ it('should glob - sync', function () {
...
 
+it('should return [] for all negative patterns - sync', function () {
+	assert.deepEqual(globby.sync(['!a.tmp', '!b.tmp']), []);
+})
+
+it('should return [] for all negative patterns - async', function () {
+	return globby(['!a.tmp', '!b.tmp']).then(function (paths) {
+		assert.deepEqual(paths, []);
+	});
+})
```
wbinnssmith: Let me know if you think these test cases are worthwhile. Wanted to hit the code path to cover the callback reference I missed.

UltCombo: :+1: 
It's odd this wasn't covered yet.

sindresorhus:
Excellent work @wbinnssmith. The code is so much nicer now :)

sindresorhus:
Published as 3.0.0.

PR #19, opened 2015-10-11
typo in Readme

bitjson:


UltCombo:
FWIW this is a typo because the [returned path](https://github.com/sindresorhus/globby/pull/19/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9L26) does not match the [actual path](https://github.com/sindresorhus/globby/pull/19/files#diff-0730bb7c2e8f9ea2438b52e419dd86c9L18).

Though, all other subdirectories are singular, I wonder if it wouldn't make more sense fixing the return value. @sindresorhus

PR #22, opened 2016-02-22
Bump glob to v7.x

shinnn:
includes a bug fix https://github.com/isaacs/node-glob/commit/0ffab1f3cbe2f3c9ec6eab21dc16c8e0d1fe3e0a

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   "dependencies": {
     "array-union": "^1.0.1",
     "arrify": "^1.0.0",
-    "glob": "^6.0.1",
+    "glob": "^7.0.0",
     "object-assign": "^4.0.1",
     "pify": "^2.0.0",
     "pinkie-promise": "^2.0.0"


UltCombo:
:+1: 
We should publish a new major version after this is merged, I think.

sindresorhus:
Nothing really useful in that release: https://github.com/isaacs/node-glob/commit/b63c6573059bf7bb1be04670b24f688612efcf71#diff-42ba1d994f4fa8c6ad17a7efae7936ccR3

`glob` seems to be on a major bumping marathon, so I'd rather wait on 8.0.0 (in a couple of weeks? :p). Major versions are expensive for users and this doesn't justify one.

Keeping this open for discovery.

shinnn:
No need to keep this open for discovery, if the next major version will be really released soon.

sindresorhus:
@shinnn That was a joke. I have no insight into when the next version will be released.

I want to keep it open or I'm guaranteed to get many useless PRs bumping the version.

shinnn:
@sindresorhus OK.

UltCombo:
Looks like node-glob's major hasn't changed for a couple months now. I believe we should upgrade and publish a major release.

UltCombo:
Rebased and updated your commit slightly: 678006bb77a6cdb14475d79fca86ba4f588e4094

PR #12, opened 2015-01-10
added ignore option for sync

vegetableman:
As discussed on isaacs/node-glob#151

PR #23, opened 2016-03-19
fix promise support for benchmarks

schnittstabil:
Benchmarks were broken since globby switched to Promises...

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -8,20 +8,16 @@ var globby = require('./');
 
 var BENCH_DIR = 'bench';
 
-function globbyToCb(globbyFn) {
-	return function (patterns, cb) {
-		setImmediate(function () {
-			globbyFn(patterns).then(cb.bind(null, null), cb);
-		});
-	};
-}
-
 var runners = [{
 	name: 'globby async (working directory)',
-	run: globbyToCb(globby)
+	run: function (patterns, cb) {
+		globby(patterns).then(cb.bind(null, null), cb);
+	}
 }, {
 	name: 'globby async (upstream/master)',
-	run: globbyToCb(globbyMaster)
+	run: function (patterns, cb) {
+		globbyMaster(patterns).then(cb.bind(null, null), cb);
+	}
 }, {
 	name: 'globby sync (working directory)',
 	run: function (patterns) {
--- a/bench.js
+++ b/bench.js
@@ -8,16 +8,20 @@ var globby = require('./');
 
 var BENCH_DIR = 'bench';
 
+function globbyToCb(globbyFn) {
+	return function (patterns, cb) {
+		setImmediate(function () {
+			globbyFn(patterns).then(cb.bind(null, null), cb);
+		});
+	};
+}
+
 var runners = [{
 	name: 'globby async (working directory)',
-	run: function (patterns, cb) {
-		setImmediate(globby, patterns, cb);
-	}
+	run: globbyToCb(globby)
 }, {
 	name: 'globby async (upstream/master)',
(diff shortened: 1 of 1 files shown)


SamVerschueren:
Might be a good idea if we extracte the promise-to-cb code in in separate library?

sindresorhus:
Can't we do the inverse and convert the non-promise APIs to promise?

schnittstabil:
@sindresorhus [matcha](https://github.com/logicalparadox/matcha) doesn't support Promises :(

schnittstabil:
@sindresorhus Is there any reason we need `setImmediate` here?

bench.js
```diff
@@ -8,16 +8,24 @@ var globby = require('./');
...
 var BENCH_DIR = 'bench';
 
+function globbyToCb(globbyFn) {
+	return function (patterns, cb) {
+		setImmediate(function () {
+			globbyFn(patterns).then(function () {
+				cb();
+			}, function (err) {
+				cb(err);
+			});
```
sindresorhus:
```js
globbyFn(patterns).then(cb.bind(null, null), cb);
```

schnittstabil: Nice :yum:

UltCombo:
@schnittstabil `setImmediate` was added to workaround a "maximum call stack size exceeded" error in the past, if I recall correctly. I believe that shouldn't happen when you are using promises, so it may be safe to remove.

UltCombo:
More specifically, `setImmediate` was introduced by #12 and the reasoning can be found at https://github.com/isaacs/node-glob/pull/151#issuecomment-69466751.

This was all the way back when globby had a callback API, before the adoption of promises. As you may know, promises are scheduled as [microtasks](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) so they are guaranteed to resolve asynchronously. I believe this should never result in a call stack size exceeded error.

So, in short, remove the `setImmediate` and run the tests, if they don't throw a call stack size exceeded error, this is good to go. :smiley:

schnittstabil:
@UltCombo Many thanks; looks like `setImmediate` is no longer needed.

bench.js
```diff
@@ -8,20 +8,16 @@ var globby = require('./');
...
-			globbyFn(patterns).then(cb.bind(null, null), cb);
-		});
-	};
-}
-
 var runners = [{
 	name: 'globby async (working directory)',
-	run: globbyToCb(globby)
+	run: function (patterns, cb) {
+		globby(patterns).then(cb.bind(), cb);
```
UltCombo: I think you still need `.bind(null, null)` (bind `null` to the callback's `this` and `err` argument), otherwise the promise's fulfillment value will be passed as the error argument to matcha's callback, isn't it?

bench.js
```diff
@@ -8,20 +8,16 @@ var globby = require('./');
...
+	}
 }, {
 	name: 'globby async (upstream/master)',
-	run: globbyToCb(globbyMaster)
+	run: function (patterns, cb) {
+		globbyMaster(patterns).then(cb.bind(), cb);
```
UltCombo: Same as [above](https://github.com/sindresorhus/globby/pull/23/commits/497c3c7324ee27d58ff36704cc6920c4d233b02a#r56760396).

bench.js
SamVerschueren: :+1:

schnittstabil: I had first to drink my coffee :sleeping:

sindresorhus:
Awesome. Thanks @schnittstabil :)

PR #26, opened 2016-03-30
Expose generateGlobTasks

schnittstabil:
I've done some clean up and unified the sync and async implementations.

The main purpose is to expose the new `generateGlobTasks` function, which I want to use in `cpy` for the `recursive` option, so that I don't need to re-implement it there.

@UltCombo: I know you have little time, but I've changed mostly code you have introduced. I would be very happy if you will find some time to review it.

@sindresorhus: Maybe `generateGlobTasks` should be extracted into its own package, but I'm not sure  how useful this would be.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,63 +6,52 @@ var glob = require('glob');
 var arrify = require('arrify');
 var pify = require('pify');
 
-function sortPatterns(patterns) {
+var globP = pify(glob, Promise).bind(glob);
+
+function isNegative(pattern) {
+	return pattern[0] === '!';
+}
+
+function generateGlobTasks(patterns, opts) {
+	var globTasks = [];
+
 	patterns = arrify(patterns);
+	opts = objectAssign({ignore: []}, opts);
 
-	var positives = [];
-	var negatives = [];
+	patterns.forEach(function (pattern, i) {
+		if (isNegative(pattern)) {
+			return;
+		}
 
-	patterns.forEach(function (pattern, index) {
-		var isNegative = pattern[0] === '!';
-		(isNegative ? negatives : positives).push({
-			index: index,
-			pattern: isNegative ? pattern.slice(1) : pattern
+		var ignore = patterns.slice(i).filter(isNegative).map(function (pattern) {
+			return pattern.slice(1);
+		});
+
+		globTasks.push({
+			pattern: pattern,
+			opts: objectAssign({}, opts, {
+				ignore: opts.ignore.concat(ignore)
+			})
 		});
 	});
 
-	return {
-		positives: positives,
-		negatives: negatives
-	};
-}
-
-function setIgnore(opts, negatives, positiveIndex) {
-	opts = objectAssign({}, opts);
-
-	var negativePatterns = negatives.filter(function (negative) {
-		return negative.index > positiveIndex;
-	}).map(function (negative) {
-		return negative.pattern;
-	});
-
-	opts.ignore = (opts.ignore || []).concat(negativePatterns);
-	return opts;
+	return globTasks;
 }
 
 module.exports = function (patterns, opts) {
-	var sortedPatterns = sortPatterns(patterns);
-	opts = opts || {};
+	var globTasks = generateGlobTasks(patterns, opts);
 
-	if (sortedPatterns.positives.length === 0) {
-		return Promise.resolve([]);
-	}
-
-	return Promise.all(sortedPatterns.positives.map(function (positive) {
-		var globOpts = setIgnore(opts, sortedPatterns.negatives, positive.index);
-		return pify(glob, Promise)(positive.pattern, globOpts);
+	return Promise.all(globTasks.map(function (task) {
+		return globP(task.pattern, task.opts);
 	})).then(function (paths) {
 		return arrayUnion.apply(null, paths);
 	});
 };
 
 module.exports.sync = function (patterns, opts) {
-	var sortedPatterns = sortPatterns(patterns);
+	var globTasks = generateGlobTasks(patterns, opts);
 
-	if (sortedPatterns.positives.length === 0) {
-		return [];
-	}
-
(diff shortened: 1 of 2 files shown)


UltCombo:
Nice, thanks for contributing.
I think I'll have time to review this late tonight.

index.js
```diff
@@ -6,63 +6,54 @@ var glob = require('glob');
...
 	});
 
-	opts.ignore = (opts.ignore || []).concat(negativePatterns);
-	return opts;
+	return globTasks;
 }
 
 module.exports = function (patterns, opts) {
-	var sortedPatterns = sortPatterns(patterns);
-	opts = opts || {};
```
UltCombo: This is no longer necessary because passing `undefined` to `objectAssign` is fine, right?

UltCombo: (I know how `Object.assign`/`objectAssign` works, just making sure your reasoning for removing this is because you're always making a shallow copy of the options object using `objectAssign` before modifying it)

UltCombo: This is covered by our tests so I guess it doesn't really matter.

index.js
```diff
@@ -6,63 +6,54 @@ var glob = require('glob');
...
+	var globTasks = generateGlobTasks(patterns, opts);
 
-	if (sortedPatterns.positives.length === 0) {
```
UltCombo: I see we no longer have this optimization, but well, the "all negative patterns" test is still passing and I guess this was an unnecessary micro-optimization after all.

test.js
```diff
@@ -70,3 +70,11 @@ it('should not mutate the options object - async', function () {
 it('should not mutate the options object - sync', function () {
 	globby.sync(['*.tmp', '!b.tmp'], Object.freeze({ignore: Object.freeze([])}));
 });
+
+it('should expose generateGlobTasks', function () {
+	var tasks = globby.generateGlobTasks(['*.tmp', '!b.tmp'], {ignore: ['c.tmp']});
+
+	assert.strictEqual(tasks.length, 1);
+	assert.strictEqual(tasks[0].pattern, '*.tmp');
+	assert.deepEqual(tasks[0].opts.ignore.sort(), ['b.tmp', 'c.tmp']);
```
UltCombo: The `.sort()` implies that the order is non-important and/or non-deterministic. I guess that may be so, but I'd normally expect the passed in `ignore` option patterns to precede the ones added by globby's implementation (generateGlobTasks).

index.js
```diff
@@ -6,63 +6,54 @@ var glob = require('glob');
...
+	patterns.forEach(function (pattern, i) {
+		if (isNegative(pattern)) {
+			return;
+		}
 
-	var negativePatterns = negatives.filter(function (negative) {
-		return negative.index > positiveIndex;
-	}).map(function (negative) {
-		return negative.pattern;
+		var ignore = patterns.slice(i).filter(isNegative).map(function (pattern) {
```
UltCombo: I was wondering if we could avoid having to iterate/filter/map all the posterior patterns for each pattern. I believe that was the motivating reason when I wrote the `sortPatterns` function to split the patterns into positive and negative groups.

I'm not sure if re-implementing that logic wouldn't overcomplicate this function too much. I guess there's not a significant perf hit unless you are passing 100's or 1000's of glob patterns, in which case this part would likely be the smallest of the bottlenecks.

UltCombo:
Excellent!

I really like our new data structures and the overall simplicity of the new implementation, it seems so much easier to follow.

I've added some comments regarding perf and the new API, though they ended up being quite monologue'ish. Please take a look at comments and add something if you would like, otherwise we can go ahead with the merge and push a minor release. :smile:

test.js
schnittstabil: At least it would be nice to detect future changes in the sequence.

index.js
schnittstabil: if `sortedPatterns.positives` is empty, we have to unnecessarily iterate a single time over the empty array (`sortedPatterns.positives.map`)…

index.js
schnittstabil: Yes, that was one of the reasons, despite merging the default option values (`{ignore: []}`).

index.js
schnittstabil: I agree, the file system is the real bottleneck. I prefer my version, because it is easier to grasp how negative patterns have an effect, that is clearly opinionated though.

If you think it is worth, I suggest instead of the
```js
var negativePatterns = negatives.filter(function (negative) {
	return negative.index > positiveIndex;
}).map(function (negative) {
	return negative.pattern;
});
```
logic, some caching strategy:
```js
var negativePatterns = negatives.slice(negativesOf[i]);
```

But I'm not sure about significant overall speed advantages.

index.js
UltCombo: In the new `generateGlobTasks` logic, in this case it would map an empty array which is a no-op (returns another empty array), which is then passed to `Promise.all` and resolves to an empty array, right? Let's avoid unnecessary logic branches, It is better this way than having to add `Promise.resolve([])` in early returns. :smiley:

index.js
UltCombo: :+1:

index.js
UltCombo: I prefer your version as well, I guess this part is not really a bottleneck.

UltCombo:
@schnittstabil Your last amended commit only changed the unit test, right?
Sorry for the delay, this is good to be merged. :smile:

UltCombo:
I've squashed and landed your PR here: 88bcb6b64a0b79f4fe0e2a5d87ef3c0027904ebd
Thank you very much! :smile_cat:
I've also documented the new function here: 9f27b1459308c149b69fc638091e7199936ca8aa


@sindresorhus Can you review the doc and publish a new minor release?

sindresorhus:
Sorry about the long delay. I've added you both to the npm package so you can publish in the future :)

PR #28, opened 2016-05-23
Share filesystem caches between tasks.

jamestalmage:
Should improve performance.

From the [`node-glob` docs](https://github.com/isaacs/node-glob#options):

> At the very least, you may pass in shared symlinks, statCache, realpathCache, and cache options, so that parallel glob operations will be sped up by sharing information about the filesystem.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,13 @@ function generateGlobTasks(patterns, opts) {
 	var globTasks = [];
 
 	patterns = arrify(patterns);
-	opts = objectAssign({ignore: []}, opts);
+	opts = objectAssign({
+		cache: Object.create(null),
+		statCache: Object.create(null),
+		realpathCache: Object.create(null),
+		symlinks: Object.create(null),
+		ignore: []
+	}, opts);
 
 	patterns.forEach(function (pattern, i) {
 		if (isNegative(pattern)) {


jamestalmage:
My only thought is that this is a breaking behavior if people are using `generateGlobTasks` to generate a set of tasks once, then reusing that task list again and again over a long period of time. You wouldn't want the cache in that scenario. Instead, people should just use `generateGlobTasks` to create a new list each time. Not sure if anyone is using it in a way that will cause a problem. Just throwing it out there.

UltCombo:
LGTM, these are good defaults for the most common use cases (globby, cpy etc.), and one can still disable the cache for the edge cases you have mentioned—perhaps this part should be documented?

/cc collaborators: Can you take a look at this PR as well?

sindresorhus:
LGTM

-

> and one can still disable the cache for the edge cases you have mentioned—perhaps this part should be documented?

👍

jamestalmage:
How would you disable the caches? Set all 5 properties to `null`? 

What if we just offered a `noCache` option that did that for you? I could go either way on this. It's more convenient, but so `niche` I'm not sure it will EVER be used.

sindresorhus:
@jamestalmage Nah, just what you said to the docs:

> people should just use generateGlobTasks to create a new list each time.

UltCombo:
> What if we just offered a `noCache` option that did that for you? I could go either way on this. It's more convenient, but so `niche` I'm not sure it will EVER be used.

I'm in the same boat. :sweat_smile: Setting 5 options to null seems cumbersome, but adding a level of indirection that not many people will use also does not seem nice.

> people should just use generateGlobTasks to create a new list each time.

Oh, nice and elegant solution! :+1:

schnittstabil:
LGTM

-

> My only thought is that this is a breaking behavior if people are using `generateGlobTasks` to generate a set of tasks once, then reusing that task list again and again over a long period of time. You wouldn't want the cache in that scenario.

If people provide caches, `generateGlobTasks` reuses them. If they don't, `generateGlobTasks` creates new cache objects __on each call(!)__ – absolutely predictable to me.

UltCombo:
@schnittstabil The point is to share the same cache object(s) across all globbing patterns passed in a single `globby()`/`globby.generateGlobTasks()` call.

schnittstabil:
@UltCombo Thanks

-

LGTM

UltCombo:
Rebased and pushed this PR here: 053f03496b3c7d7fe857fb7dfd0530d8b5512705
Documented `generateGlobTasks`'s caching behavior here: f2bd72b126f45926340df297d7a5c2b5f61f2c6b
Sorry for the delay, this almost fell off my radar.

By the way, should this be considered a breaking change (major release)? It is technically a breaking change, but only for very rare edge cases involving a quite specialized low-level method, as [commented above](https://github.com/sindresorhus/globby/pull/28#issuecomment-221050569). I believe it should not break anything in the wild. WDYT?

jamestalmage:
It's a breaking change on a documented method, so I would do a major bump. Especially because, if someone **is** affected - it's likely going to be really confusing and hard for them to debug.

UltCombo:
Yep, better safe than sorry. I guess we can couple this with a node-glob update (#22) and release a new major. :smile:

UltCombo:
Released `globby@5.0.0`! :tada:

@sindresorhus BTW I could not find a changelog and it looks like the repository was not using GitHub Releases either. Which approach do you usually prefer?
For now, I've made a [GitHub release](https://github.com/sindresorhus/globby/releases/tag/v5.0.0) to document the improvements and breaking changes.

schnittstabil:
@UltCombo See https://github.com/sindresorhus/del/issues/36#issuecomment-142647132 and https://github.com/sindresorhus/del/issues/36#issuecomment-142649019  :stuck_out_tongue_winking_eye:

sindresorhus:
GitHub Releases.

Thanks for doing a release @UltCombo :)

PR #29, opened 2016-06-08
Use AVA instead of mocha

kevva:


SamVerschueren:
Remove all the `should` words.

sindresorhus:
❤️

test.js
```diff
@@ -1,80 +1,71 @@
-'use strict';
-var assert = require('assert');
-var fs = require('fs');
-var globby = require('./');
+import fs from 'fs';
+import test from 'ava';
+import m from './';
```
UltCombo: I'm 👎  on renaming `globby` -> `m`. Users often look for usage examples in the tests file; it should be optimized for readability.
Any strong reasons to prefer this other naming pattern?

package.json
```diff
@@ -60,17 +60,11 @@
     "pinkie-promise": "^2.0.0"
   },
   "devDependencies": {
+    "ava": "*",
```
UltCombo: Perhaps use a proper semver range (`^X.X.X`)? So the build wouldn't break when breaking changes are released in a major version.

sindresorhus: I do this on all my repos. I'm ok with stuff breaking on my own repos so I can catch bugs early and prevent it to happen for my users.

UltCombo: I'm not sure I understand. If you depend on a version range such as `^1.2.3`, then you can still test all releases that are supposedly backwards-compatible with the version that you depend on (e.g. `1.4.6`, `1.20.99`).
When you do release a version `2.0.0`, that means there are intentional breaking changes and users are expected to update existing code before upgrading the dependency.

As far as I can see, `*` just completely misses the point of semver for no good reason.

sindresorhus: AVA is pre-1.0.0, so `^0.1.0` doesn't match `0.2.0`.

UltCombo: Ohh makes sense, I forgot to check the latest version. 😅

PR #31, opened 2016-07-10
perform validation on the patterns parameter and error accordingly

### LINKED ISSUE #30: passing in improper pattern values throws instead of rejecting the promise
There is no validation on the patterns. It is assumed that they can be used, and the code throws an error if they can't instead of rejecting the promise and triggering the `catch`.

Example:

```javascript
globby([null]).then(...).catch(function(err) {
  // expect error to happen
});
```

This throws [here](https://github.com/sindresorhus/globby/blob/053f03496b3c7d7fe857fb7dfd0530d8b5512705/index.js#L12), where the expectation is that it can read the pattern as a string when it is not a string.

catdad:
Fixes #30.

I don't know what you like on the actual language of the error message, so let me know prefer something else.

### DIFF
--- a/index.js
+++ b/index.js
@@ -23,11 +23,15 @@ function validatePatternsArray(patterns) {
 	}
 
 	if (!Array.isArray(patterns) || !patterns.every(isString)) {
-		return new TypeError('patterns must be a string or an array of strings');
+		throw new TypeError('patterns must be a string or an array of strings');
 	}
 }
 
 function generateGlobTasks(patterns, opts) {
+	// make sure input is correct, so that we can safely parse
+	// it in this task
+	validatePatternsArray(patterns);
+
 	var globTasks = [];
 
 	patterns = arrify(patterns);
@@ -60,13 +64,13 @@ function generateGlobTasks(patterns, opts) {
 }
 
 module.exports = function (patterns, opts) {
-	var validationError = validatePatternsArray(patterns);
+	var globTasks;
 
-	if (validationError) {
-		return Promise.reject(validationError);
+	try {
+		globTasks = generateGlobTasks(patterns, opts);
+	} catch (err) {
+		return Promise.reject(err);
 	}
-
-	var globTasks = generateGlobTasks(patterns, opts);
 
 	return Promise.all(globTasks.map(function (task) {
 		return globP(task.pattern, task.opts);
@@ -76,12 +80,6 @@ module.exports = function (patterns, opts) {
 };
 
 module.exports.sync = function (patterns, opts) {
-	var validationError = validatePatternsArray(patterns);
-
-	if (validationError) {
-		throw validationError;
-	}
-
 	var globTasks = generateGlobTasks(patterns, opts);
 
 	return globTasks.reduce(function (matches, task) {
--- a/index.js
+++ b/index.js
@@ -3,7 +3,6 @@ var Promise = require('pinkie-promise');
 var arrayUnion = require('array-union');
 var objectAssign = require('object-assign');
 var glob = require('glob');
-var arrify = require('arrify');
 var pify = require('pify');
 
 var globP = pify(glob, Promise).bind(glob);
@@ -17,24 +16,17 @@ function isString(value) {
 }
 
 function assertPatternsInput(patterns) {
-	if (isString(patterns)) {
-		// bail early, the rest will easier this way
-		return;
-	}
-
-	if (!Array.isArray(patterns) || !patterns.every(isString)) {
+	if (!patterns.every(isString)) {
 		throw new TypeError('patterns must be a string or an array of strings');
 	}
 }
 
 function generateGlobTasks(patterns, opts) {
-	// make sure input is correct, so that we can safely parse
-	// it in this task
+	patterns = Array.isArray(patterns) ? patterns : [patterns];
 	assertPatternsInput(patterns);
 
 	var globTasks = [];
 
-	patterns = arrify(patterns);
 	opts = objectAssign({
 		cache: Object.create(null),
 		statCache: Object.create(null),
--- a/index.js
+++ b/index.js
@@ -9,7 +9,22 @@ var pify = require('pify');
 var globP = pify(glob, Promise).bind(glob);
 
 function isNegative(pattern) {
-	return pattern && pattern[0] === '!';
+	return pattern[0] === '!';
+}
+
+function isString(value) {
+	return typeof value === 'string';
+}
+
+function validatePatternsArray(patterns) {
+	if (isString(patterns)) {
+		// bail early, the rest will easier this way
+		return;
+	}
+
+	if (!Array.isArray(patterns) || !patterns.every(isString)) {
+		return new TypeError('patterns must be a string or an array of strings');
+	}
 }
 
 function generateGlobTasks(patterns, opts) {
@@ -45,6 +60,12 @@ function generateGlobTasks(patterns, opts) {
 }
 
 module.exports = function (patterns, opts) {
+	var validationError = validatePatternsArray(patterns);
+
+	if (validationError) {
+		return Promise.reject(validationError);
+	}
+
 	var globTasks = generateGlobTasks(patterns, opts);
 
 	return Promise.all(globTasks.map(function (task) {
@@ -55,6 +76,12 @@ module.exports = function (patterns, opts) {
 };
 
 module.exports.sync = function (patterns, opts) {
+	var validationError = validatePatternsArray(patterns);
+
+	if (validationError) {
+		throw validationError;
+	}
+
 	var globTasks = generateGlobTasks(patterns, opts);
 
 	return globTasks.reduce(function (matches, task) {
--- a/index.js
+++ b/index.js
@@ -22,14 +22,8 @@ function validatePatternsArray(patterns) {
 		return;
 	}
 
-	var err = new TypeError('patterns must be a string or an array of strings');
-
-	if (!Array.isArray(patterns)) {
-		return err;
-	}
-
-	if (patterns.filter(isString).length !== patterns.length) {
-		return err;
+	if (!Array.isArray(patterns) || !patterns.every(isString)) {
+		return new TypeError('patterns must be a string or an array of strings');
 	}
 }
 
--- a/index.js
+++ b/index.js
@@ -22,14 +22,14 @@ function validatePatternsArray(patterns) {
 		return;
 	}
 
+	var err = new TypeError('patterns must be a string or an array of strings');
+
 	if (!Array.isArray(patterns)) {
-		return new TypeError('patterns must be a string or an array of strings');
+		return err;
 	}
 
-	var stringPatterns = patterns.filter(isString);
-
-	if (stringPatterns.length !== patterns.length) {
-		return new TypeError('all pattners must be strings');
+	if (patterns.filter(isString).length !== patterns.length) {
+		return err;
 	}
 }
 
--- a/index.js
+++ b/index.js
@@ -12,6 +12,27 @@ function isNegative(pattern) {
 	return pattern[0] === '!';
 }
 
+function isString(value) {
+	return typeof value === 'string';
+}
+
+function validatePatternsArray(patterns) {
+	if (isString(patterns)) {
+		// bail early, the rest will easier this way
+		return;
+	}
+
+	if (!Array.isArray(patterns)) {
+		return new TypeError('patterns must be a string or an array of strings');
+	}
+
+	var stringPatterns = patterns.filter(isString);
+
+	if (stringPatterns.length !== patterns.length) {
+		return new TypeError('all pattners must be strings');
+	}
+}
+
 function generateGlobTasks(patterns, opts) {
 	var globTasks = [];
 
@@ -45,6 +66,12 @@ function generateGlobTasks(patterns, opts) {
 }
 
 module.exports = function (patterns, opts) {
+	var validationError = validatePatternsArray(patterns);
+
+	if (validationError) {
+		return Promise.reject(validationError);
+	}
+
 	var globTasks = generateGlobTasks(patterns, opts);
 
 	return Promise.all(globTasks.map(function (task) {
@@ -55,6 +82,12 @@ module.exports = function (patterns, opts) {
 };
 
 module.exports.sync = function (patterns, opts) {
+	var validationError = validatePatternsArray(patterns);
+
+	if (validationError) {
+		throw validationError;
+	}
+
 	var globTasks = generateGlobTasks(patterns, opts);
 
 	return globTasks.reduce(function (matches, task) {
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@ function isString(value) {
 	return typeof value === 'string';
 }
 
-function validatePatternsArray(patterns) {
+function assertPatternsInput(patterns) {
 	if (isString(patterns)) {
 		// bail early, the rest will easier this way
 		return;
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -12,6 +12,27 @@ function isNegative(pattern) {
 	return pattern[0] === '!';
 }
 
+function isString(value) {
+	return typeof value === 'string';
+}
```
sindresorhus: No need for a function for this. Can just be inline.

index.js
```diff
@@ -12,6 +12,27 @@ function isNegative(pattern) {
...
+
+function validatePatternsArray(patterns) {
+	if (isString(patterns)) {
+		// bail early, the rest will easier this way
+		return;
```
sindresorhus: This needs to return an error too.

index.js
```diff
@@ -12,6 +12,27 @@ function isNegative(pattern) {
...
+	}
+
+	var err = new TypeError('patterns must be a string or an array of strings');
+
+	if (!Array.isArray(patterns)) {
+		return err;
+	}
+
+	if (patterns.filter(isString).length !== patterns.length) {
```
sindresorhus: `patterns.every(isString)`

You can also move both checks into one `if` statement.

catdad: I always forget about that. Done, thanks.

sindresorhus:
LGTM

@UltCombo ?

UltCombo:
Couple suggestions:

- It may be a good idea to perform the validation inside the `generateGlobTasks` function, so that the low-level `globby.generateGlobTasks` API also benefits from it and we don't need to explicitly call the validation in both sync and async versions (will need to catch the exception in the promise version if throwing from inside `generateGlobTasks`, though).

- Another approach that could be explored is to just fix the `isNegative` function to not throw an error when receiving a `null`/`undefined` pattern (e.g. `return pattern && pattern[0] === '!';`), then pass the glob tasks to node-glob as usual and catch a possible exception if any. This seems more future-proof and keeps our API inline with node-glob, also it saves a bunch of code and we don't need to create a new error message.

catdad:
@UltCombo Thanks for the feedback. Here are the problems I am looking at now:

Handling the error in `globby.generateGlobTasks` does need to happen. I generally don't like to `try/catch` when I don't have to, but that is more of a personal preference.

As for not validating at all... I guess I should have written more tests. You are right, for `null`/`undefined`, just letting node-glob error is fine. They have similar error messages already. However, those are not the only possible values. For example, `[['!']]` seems to work fine instead of telling you that you just did a stupid thing. [Surprisingly relevant, and funny.](https://twitter.com/sempf/status/514473420277694465)

Anyway, I am looking into how to correctly handle with delegtion to node-glob itself, while still providing a good validation experience, and will likely write a bunch more tests for this. In general, I have found that trying to save code by cutting out validation is an unworthy cause.

UltCombo:
Thanks for looking into this!

> I generally don't like to try/catch when I don't have to, but that is more of a personal preference.

I personally don't like it either, but from a quick glance I couldn't think of a more appropriate approach—it is generally an anti-pattern to return error instances.

> In general, I have found that trying to save code by cutting out validation is an unworthy cause.

I didn't actually mean to cut out the validation, but rather delegate it to node-glob in order to ensure that we are not over-constraining the input.

---

> For example, `[['!']]` seems to work fine instead of telling you that you just did a stupid thing. Surprisingly relevant, and funny.

Well, it is virtually impossible to defend against every single possible misuse.

catdad:
Okay, so I ended up with these tests (all pass), which I think is pretty ugly... especially for the cases where sync and async generate different errors.

```javascript
// false positive tests
[
	null,
	[['!']],
	{0: '!'}
].forEach(v => {
	const name = 'successfully gets empty array for pattern value: ' + (JSON.stringify(v) || v.toString());

	test(name + ' - async', async t => {
		t.deepEqual(await m(v), []);
	});

	test(name, async t => {
		t.deepEqual(m.sync(v), []);
	});
});

// rejected for being an invalid pattern
[
	{},
	true,
	[true],
	function () {},
	[function () {}]
].forEach(v => {
	const name = 'errors for invalid pattern value: ' + (JSON.stringify(v) || v.toString());

	test(name + ' - async', async t => {
		t.throws(m(v), TypeError);
		t.throws(m(v), 'glob pattern string required');
	});

	test(name, t => {
		t.throws(() => m.sync(v), TypeError);
		t.throws(() => m.sync(v), 'glob pattern string required');
	});
});

// rejected, but have a different error to async and sync
[
	[null],
	false,
	[false]
].forEach(v => {
	const name = 'errors for invalid pattern value: ' + (JSON.stringify(v) || v.toString());

	test(name + ' - async', async t => {
		t.throws(m(v), TypeError);
		t.throws(m(v), 'glob pattern string required');
	});

	test(name, t => {
		t.throws(() => m.sync(v), Error);
		t.throws(() => m.sync(v), 'must provide pattern');
	});
});
```

Do you think this is the right way to go? It accomplishes the task by rejecting the promise (or erroring in sync) correctly, but it does cause globby to use node-glob incorrectly for a few cases, and is a bit ugly with the different sync and async errors.

> Well, it is virtually impossible to defend against every single possible misuse.

You do this by validating that the assumptions you are making are actually true (such as, don't parse something as a string without checking that it is a string).

UltCombo:
> but it does cause globby to use node-glob incorrectly for a few cases, and is a bit ugly with the different sync and async errors.

Are those errors coming from node-glob? If so, I guess it is not really a problem.

> You do this by validating that the assumptions you are making are actually true (such as, don't parse something as a string without checking that it is a string).

Not sure if we are on the same page. I'm talking about the big picture—the whole JavaScript ecosystem is based on implicit contracts and protocols. You can see that several libraries will fail silently or implicitly cast the arguments to the expected type. And, sadly, changing this kind of behavior is often a breaking change as popular libraries always have users that (often unknowingly) depend on this behavior to work.

catdad:
Yeah, all of those errors are coming from node-glob. I'll go ahead and push it to this PR then.

As for the more philosophical question, not sure that others' bad behavior is a good argument for more bad behavior. Because you can make implicit contracts doesn't mean that you should or that the language is based on it. It just means you are responsible for the errors that you make.

UltCombo:
> Because you can make implicit contracts doesn't mean that you should or that the language is based on it.

The language's lack of type safety and pattern matching forces us to have such contracts. A good part of the language performs type casting (which only becomes explicit when reading the spec.), but in any case, I meant that the ecosystem (userland packages and libraries) often rely on this language aspect, as it is an extra effort going against the tide to have it otherwise. But well, I guess this philosophical discussion won't really get us anywhere, so let's move on.

On a side note, I also greatly value brevity. Chopping away code === smaller cognitive load, less code to maintain, less surface area for bugs.

> not sure that others' bad behavior is a good argument for more bad behavior.

I fully agree, and I dislike having these API inconsistencies just as much as you do. And my bad, I'm not opposed to improving the validation—as I mentioned, it was an approach worth exploring. Thanks for exploring it and finding all these inconsistencies.

My previous comment merely mentions that our current implicit contract is to pass the patterns as-is to node-glob (with an unexpected exceptional case for `null`/`undefined` patterns). I'm open to improving our validation, but it breaks our current implicit contract and thus should be released as a breaking change.

UltCombo:
Just an addendum.

Still in the contracts area, I believe you cannot "make" an implicit contract—the fact it is implicit means that nothing explicit has been made about it. 😉 

And in case my previous comment seems contradictory: I do value brevity, but I also value user-friendly APIs that help users fixing errors. Software development is a chaos area constantly seeking the perfect balance.

catdad:
I'm sorry, I wanted to take this seriously, but I just cannot read that much text with the new GitHub fonts.

Anyway, all the changes are available here now. Let me know if you have any other thoughts.

catdad:
But one last one, because I skimmed it and it now bothers me. Less code does not mean less cognitive load, and my original code was a great example of that.

By adding validation to make sure the parameter is a string, you know exactly what the behavior of your module is: it errors for anything other than a string (the only parameter that is actually "correct"), and therefore, you have to only think about strings from that point on.

In the current version, where we have less code, but pass any value to node-glob, there is actually more cognitive load. Can you tell me off the top of your head what behavior is expected when I pass in `null`? What about `NaN`? What about `true`? No? It's because the code is now much more difficult to think about, even though there is less of it. There are many more unexpected things. And in order to know what that behavior is, you have to do many more tests. Same applies to bugs and maintaining code.  You have to understand the behavior before you can maintain the behavior, and maintaining large unknown behavior is harder than maintaining smaller specific behavior.

UltCombo:
I normally don't think about unsupported/undocumented inputs; it is undefined behavior and may fail in unexpected ways. But ok, we can agree to disagree on these theoretical issues.

Please read the second part of the my comment above. We can reduce API inconsistencies and improve user-friendliness, which outweigh brevity and other theoretical aspects, therefore I'm open to adding a validation layer. I've already mentioned this in the comments above.

catdad:
I don't think that is necessary at this point. I set out to get a meaningful error and to be able to use the Promise as intended, and that works now. Everything else should be as you prefer it, since it is your module after all.

UltCombo:
I don't have a strong preference at this point, but I believe user-friendliness trumps other attributes.

See the two last paragraphs in the reply [above](https://github.com/sindresorhus/globby/pull/31#issuecomment-231755072), if you care. I'm ending towards reverting back to your first proposed approach, but I'm extremely short of time tonight.

Meanwhile, @sindresorhus care to provide an opinion here?

sindresorhus:
I prefer explicitness. I realize JS is inherently implicit, but that doesn't mean we can't do better. I was fine with having some validation in this module. I agree that also makes the code more readable as you can clearly see what the type constraints of the various inputs. I don't agree validation of already documented behavior is a breaking change. Semver defines breaking changes as changes to the public API.

catdad:
Thanks for the input. I have reverted the last commit, so that validation is again done in `globby`, and I also moved the validation to the shared `generateGlobTasks`, so that all methods will not do the same validation.

UltCombo:
> Semver defines breaking changes as changes to the public API.

Exactly, all of globby's entry points are part of the public API and were accepting some non-documented types until now. Now, I believe you may have meant "breaking changes as changes to the ***documented*** public API", which is a big gray zone. You may consider the acceptance of non-documented types as a bug, but if there are people (un)knowingly relying on this bug then it is also breaking change.

@sindresorhus I'm aware you already had a long discussion about this very topic with @bevacqua on this [Twitter thread](https://twitter.com/nzgb/status/652215907742285825), and it seems a compromise has never been reached. Developers tend to think in absolutes, it makes decision making far easier. I get that.

To be clear, I still mostly agree with your reasoning, and as far as I can see it is plainly impossible to predict every single possible misuse that could be considered a breaking change, so we have to stick with keeping compatibility with the documented API. However, more recently I've been trying to reach a compromise: if I can see a change that is (at least) slightly likely to break other people's builds, I'd prefer to play in the safe side and release it as a breaking change. Honestly, the very worst experience in Node.js development is when such such small "breaking" changes slip by in a package's patch/minor version which is used somewhere among another 500 project dependencies and make the whole build crash. Then I have to spend hours hunting down what some package's author considered an unlikely breaking change, which quite often is actually an unlikely and hard to find bug, buried deep down in the dependencies tree.

So, I'd prefer to play it safe and save people from going all the way down the rabbit hole in such cases. But anyway, it is your call.

@catdad Thanks for sticking around and updating the PR! :smile:

index.js
```diff
@@ -12,7 +12,26 @@ function isNegative(pattern) {
...
+}
+
 function generateGlobTasks(patterns, opts) {
+	// make sure input is correct, so that we can safely parse
+	// it in this task
+	validatePatternsArray(patterns);
+
 	var globTasks = [];
 
 	patterns = arrify(patterns);
```
schnittstabil: @catdad What do you think about, changing `validatePatternsArray` to `assertPatternsArray` (or similar) and:

```js
function generateGlobTasks(patterns, opts) {
  	var globTasks = [];
  
  	patterns = arrify(patterns);
	assertPatternsArray(patterns);
```

The benefits I see:
1. [assert](https://nodejs.org/api/assert.html) is a well-known term, known to throw errors
2. There is no need anymore to check `isString(patterns)` in `validatePatternsArray`/`assertPatternsArray`

schnittstabil: Aww, I've forgotten:
> arrify:
> Supplying null or undefined results in an empty array.

This may not suffice your needs…

catdad: That last note is exactly why I don't want to do `arrify` before validation. It will cause some invalid input to be ignored, and to me, that is weird behavior. I can do the rest a bit later today.

catdad:
Since this conversation is happening, I would like to mention that the public API is not only the methods available to call, but also the parameters of those methods, the return values (and values provided to callbacks/Promises), and the errors that those methods produce. If those things are not documented, it is not "undocumented behavior", but rather a bug. If a method is documented to take only a string or array of strings, then passing in any other value and having the method continue is also a bug, and not "incorrect usage of the API". Generally, writing this behavior and complete unit tests for this behavior is the a great way to avoid releasing minor breaking changes that cause huge problems.

I find that the best way to deal with this is to have comprehensive unit tests, and where appropriate, contract tests for your public API. In fact, I found this globby bug when I was trying to unit test my error handling code (since you must write error handling code).

Anyway, I do like the name @schnittstabil suggested, so I have updated accordingly. I also added more comprehensive tests for different value types, which will now fail if arrify is moved ahead of validation (or similar changes happen).

sindresorhus:
I guess we can do a major release to be nice, but then we're inconveniencing everyone using the API correctly in favor of not breaking the few that does not.

index.js
```diff
@@ -12,7 +12,26 @@ function isNegative(pattern) {
...
+}
+
 function generateGlobTasks(patterns, opts) {
+	// make sure input is correct, so that we can safely parse
+	// it in this task
+	assertPatternsInput(patterns);
+
 	var globTasks = [];
 
 	patterns = arrify(patterns);
```
schnittstabil: @catdad If `null` and `undefined` are considered as errors, then I don't see any benefit of using [`arrify`](https://github.com/sindresorhus/arrify/blob/v1.0.1/index.js) at all:

```js
function generateGlobTasks(patterns, opts) {
  	var globTasks = [];
  
  	patterns = Array.isArray(patterns) ? patterns : [patterns];
	assertPatternsInput(patterns);
```

(and `assertPatternsInput` could be simplified, as already suggested)

Furthermore, as we use (self-explanatory) speaking names, personally, I see no advantage of comments and would remove them.

UltCombo: Excellent points, @schnittstabil !

@catdad Would you like to apply these changes or do you prefer us to do so?

catdad: I am not actually sure what is being suggested here. Are you saying that `arrify` should be used before validation, and the values `null` and `undefined` should be allowed through?

The problem I have there is that `null` and `undefined` would result in a successful operation returning `[]`, but `[null]` and `[undefined]` will result in a `TypeError`. The type for patters is a string or array of strings. `null` and `undefined` are valid types with are not a string or array of strings. Yes, I agree that it is less code, but it is also more unexpected behavior.

UltCombo: @catdad Please see the code sample that @schnittstabil posted above. I believe he suggested to remove the `arrify` dependency and manually array-ify the the patterns:

```js
patterns = Array.isArray(patterns) ? patterns : [patterns];
```

Before passing them to `assertPatternsInput`. Then you can remove the `isString(patterns)` and `!Array.isArray(patterns)` checks, just a `patterns.every(isString)` inside `assertPatternsInput` will suffice.

catdad: Sorry, I am slow.

It's been committed and tests are running now.

sindresorhus:
Awesome! Looks good. Thanks @catdad :)

UltCombo:
:tada: v6.0.0 nice!
Thanks @catdad! Awesome work everyone! :smiley_cat:

PR #34, opened 2016-10-16
Support negated ignores

JonnyBurger:
This PR changes the behaviour so that entries in the `ignore` options that start with a `!` don't ignored. This is useful if you want to pass a list from `.gitignore` to the `ignore` option (as you do in `xo`).

<hr>

Example file structure:

```
├── a.js
└─── b
    ├── c.js
    └── d.js
```

Example `.gitignore`

```
b/*
!b/c.js
```

Computing which files would be added to Git:

``` js
await globby(`**/*`, {ignore: ['b/*', '!b/c.js']})
// => before: []
// => after: ['a.js', 'b/c.js']
```

Related `xo` issue: https://github.com/sindresorhus/xo/issues/154

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,7 +46,7 @@ function generateGlobTasks(patterns, opts) {
 		ignore = opts.ignore.concat(ignore);
 
 		// Convert negated ignores into positive patterns
-		ignore.forEach(function(ign) {
+		ignore.forEach(function (ign) {
 			if (!isNegative(ign)) {
 				return;
 			}
@@ -57,8 +57,8 @@ function generateGlobTasks(patterns, opts) {
 		});
 
 		// Remove negated ignores from array
-		ignore = ignore.filter(function(ign) {
-			return !isNegative(ign)
+		ignore = ignore.filter(function (ign) {
+			return !isNegative(ign);
 		});
 
 		globTasks.push({
--- a/index.js
+++ b/index.js
@@ -78,6 +78,7 @@ module.exports = function (patterns, opts) {
 	} catch (err) {
 		return Promise.reject(err);
 	}
+
 	return Promise.all(globTasks.map(function (task) {
 		return globP(task.pattern, task.opts);
 	})).then(function (paths) {
--- a/index.js
+++ b/index.js
@@ -43,11 +43,26 @@ function generateGlobTasks(patterns, opts) {
 		var ignore = patterns.slice(i).filter(isNegative).map(function (pattern) {
 			return pattern.slice(1);
 		});
+		ignore = opts.ignore.concat(ignore);
+
+		// Convert negated ignores into positive patterns
+		ignore.forEach(ign => {
+			if (!isNegative(ign)) {
+				return;
+			}
+			globTasks.push({
+				pattern: ign.slice(1),
+				opts: objectAssign({}, opts, {ignore: []})
+			});
+		});
+
+		// Remove negated ignores from array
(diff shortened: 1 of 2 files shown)


JonnyBurger:
Changed the arrow functions to regular `function()`'s, so node 0.10 should be supported. The build will still fail because `xo` has semver `*` in `package.json` and does not anymore support node 0.10, but it doesn't have to do with my changes.

schnittstabil:
@JonnyBurger [As far as I can see](https://travis-ci.org/sindresorhus/globby/jobs/168093386), the build fails because of:
```
  index.js:49:26
  ✖  49:26  Missing space before function parentheses.  space-before-function-paren
  ✖  60:34  Missing space before function parentheses.  space-before-function-paren
  ✖  61:27  Missing semicolon.    
```

index.js
```diff
@@ -43,11 +43,28 @@ function generateGlobTasks(patterns, opts) {
 		var ignore = patterns.slice(i).filter(isNegative).map(function (pattern) {
 			return pattern.slice(1);
 		});
+		ignore = opts.ignore.concat(ignore);
+
+		// Convert negated ignores into positive patterns
+		ignore.forEach(function(ign) {
```
schnittstabil:
```diff
- function(ign)
+ function (ign)
```

index.js
```diff
@@ -43,11 +43,28 @@ function generateGlobTasks(patterns, opts) {
...
+				return;
+			}
+			globTasks.push({
+				pattern: ign.slice(1),
+				opts: objectAssign({}, opts, {ignore: []})
+			});
+		});
+
+		// Remove negated ignores from array
+		ignore = ignore.filter(function(ign) {
```
schnittstabil:
```diff
- function(ign)
+ function (ign)
```

index.js
```diff
@@ -43,11 +43,28 @@ function generateGlobTasks(patterns, opts) {
...
+			return !isNegative(ign)
```
schnittstabil:
```diff
- return !isNegative(ign)
+ return !isNegative(ign);
```

JonnyBurger:
Thanks. Fixed.

schnittstabil:
To be honest, I'm not sure about this. Okay, `.gitignore` and `node-glob` interpreting negation patterns differently, but this does not necessarily mean `node-glob`/`globby` interpret these wrongly.
– Can't we do this in `xo`? Or even better, with some *`gitignore-to-glob(task)`* package?
Especially, because there can be more than one `.gitignore` file, and it's meaning depends on its path…

index.js
```diff
@@ -43,11 +43,28 @@ function generateGlobTasks(patterns, opts) {
...
+		ignore = opts.ignore.concat(ignore);
+
+		// Convert negated ignores into positive patterns
+		ignore.forEach(function (ign) {
+			if (!isNegative(ign)) {
+				return;
+			}
+			globTasks.push({
+				pattern: ign.slice(1),
+				opts: objectAssign({}, opts, {ignore: []})
```
schnittstabil: `{ignore: []}` doesn't look right – what about subsequent ignore patterns?
E.g. `globby(…, {ignore: ['!*.tmp', 'b.tmp']}`.

sindresorhus:
I agree, this should be fixed in XO. It doesn't make much sense in general to pass negated patterns to ignore.

@JonnyBurger Sorry, for this very very late reply. Would you be up for doing a pull request on XO instead?

PR #36, opened 2016-10-26
expose `.hasMagic()`

### LINKED ISSUE #35: Expose glob.hasMagic
Would you mind exposing the `hasMagic` function in globby.

https://github.com/isaacs/node-glob#globhasmagicpattern-options

Moeriki:
Fixes #35.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -56,6 +56,12 @@ Type: `Object`
 
 See the `node-glob` [options](https://github.com/isaacs/node-glob#options).
 
+### globby.hasMagic(patterns, [options])
+
+Returns `true` if there are any special glob characters in an array of patterns.
+
+See `node-glob` hasMagic(https://github.com/isaacs/node-glob#globhasmagicpattern-options).
+
 
 ## Globbing patterns
 
--- a/index.js
+++ b/index.js
@@ -21,8 +21,12 @@ function assertPatternsInput(patterns) {
 	}
 }
 
+function castArray(value) {
+	return Array.isArray(value) ? value : [value];
+}
+
 function generateGlobTasks(patterns, opts) {
-	patterns = Array.isArray(patterns) ? patterns : [patterns];
+	patterns = castArray(patterns);
 	assertPatternsInput(patterns);
 
 	var globTasks = [];
@@ -80,3 +84,10 @@ module.exports.sync = function (patterns, opts) {
 };
 
 module.exports.generateGlobTasks = generateGlobTasks;
+
+module.exports.hasMagic = function (patterns, opts) {
+	patterns = castArray(patterns);
+	return patterns.some(function (pattern) {
+		return glob.hasMagic(pattern, opts);
+	});
+};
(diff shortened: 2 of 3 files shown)


readme.md
```diff
@@ -44,6 +44,12 @@ Returns an array of objects in the format `{ pattern: string, opts: Object }`, w
 
 Note that you should avoid running the same tasks multiple times as they contain a file system cache. Instead, run this method each time to ensure file system changes are taken into consideration.
 
+### globby.hasMagic(pattern, [options]);
```
schnittstabil: Personally, I'm unwilling to keep documentation in sync with `node-globs`. Furthermore, it smells like plagiarism. As we don't modify behavior, I would suggest:
> ### globby.hasMagic
> See `node-glob` [hasMagic](https://github.com/isaacs/node-glob#globhasmagicpattern-options).

Moeriki: Yes that would be much better.

Wasn't even sure if documenting it, was needed at all. Since the goal `Extends glob with support for multiple patterns and exposes a Promise API` already implied it.

schnittstabil: Please note my https://github.com/sindresorhus/globby/issues/35#issuecomment-256328382, maybe we should modify the signature–sorry for the clutter.

Moeriki:
Pull request updated.

* hasMagic documentation references node-glob
* hasMagic can take an array of paths and returns `true` if any hasMagic

readme.md
schnittstabil: Please remove the `;` and change `pattern` to `patterns`.

schnittstabil:
@sindresorhus lgtm, I've manually verified v0.10 and v0.12. I'm willing to es2015ify before the next release, to drop support for v0.10 and v0.12.

@kevva, @UltCombo What do you think?

UltCombo:
LGTM.
Nit: perhaps could use [array-ify](https://github.com/stevemao/array-ify) instead of creating a `castArray` function.

sindresorhus:
> Nit: perhaps could use array-ify instead of creating a castArray function.

Nah. Just leave it. Can be simply `Array.of()` when we target Node.j 4 in the next major version.

readme.md
```diff
@@ -44,6 +44,10 @@ Returns an array of objects in the format `{ pattern: string, opts: Object }`, w
 
 Note that you should avoid running the same tasks multiple times as they contain a file system cache. Instead, run this method each time to ensure file system changes are taken into consideration.
 
+### globby.hasMagic(patterns, [options])
+
+See `node-glob` [hasMagic](https://github.com/isaacs/node-glob#globhasmagicpattern-options).
```
sindresorhus: Should be made clearer that it differentiates by supporting multiple patterns.

UltCombo:
> Nah. Can just be inlined to `[].concat(patterns)`.

That allocates two extra array instances which may be unnecessary. I'm not sure if V8 is smart enough to optimize that, but I guess it doesn't really matter. Your call.

PR #46, opened 2017-07-04
Add `expandDirectories` option and set `nodir` to `true` by default

### LINKED ISSUE #33: Option to only return files
Currently, `globby` or rather `glob` returns both directories and files. You can turn off returning directories using the `nodir: true` glob option, but that's not what you want. You'd rather want to expand directories into files. That's what this option should do. So `globby.sync('foo')` where `foo` is a directory would be expanded into all the files in that directory that doesn't match any of the ignore patterns. Same with `globby.sync('foo/**')`. There also needs to be a way to specify an array of extensions. In many cases you wouldn't want the expansion to happen for all files, but, for example, just `.js` files. It should override the `nodir` option if specified by the user.

This would be useful and simplify code in AVA, XO, cpy, imagemin-cli, and other tools expecting files. For example: https://github.com/avajs/ava-files/blob/master/index.js#L212-L271

@UltCombo @schnittstabil @jamestalmage @SamVerschueren @novemberborn Thoughts?

kevva:
Noticed a pretty large performance regression so might have to fix that first before merging. Would've liked to to implement the login in `generateGlobTasks` but we'd need to introduce `generateGlobTasksSync` in that case.

Fixes #33.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   ],
   "dependencies": {
     "array-union": "^1.0.1",
-    "dir-glob": "kevva/dir-glob#files",
+    "dir-glob": "^2.0.0",
     "glob": "^7.1.2",
     "pify": "^2.3.0"
   },
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   ],
   "dependencies": {
     "array-union": "^1.0.1",
+    "dir-glob": "kevva/dir-glob#files",
     "glob": "^7.1.2",
     "pify": "^2.3.0"
   },
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 const arrayUnion = require('array-union');
 const glob = require('glob');
 const pify = require('pify');
+const dirGlob = require('dir-glob');
 
 const globP = pify(glob);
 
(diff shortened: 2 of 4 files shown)


kevva:
Ok, think I've nailed the performance regressions to the `nodir` option being `true`, and the `isDirectory` call here https://github.com/kevva/dir-glob/blob/0a3b078f088836d9584510d1c735875dac9dc98b/index.js#L45.

package.json
```diff
@@ -53,6 +53,7 @@
   ],
   "dependencies": {
     "array-union": "^1.0.1",
+    "dir-glob": "kevva/dir-glob#files",
```
pkuczynski: Temporary?

kevva: Yes, that branch hasn't been merged to `master` yet.

sindresorhus:
> I'm wondering if we maybe should check if it's a glob using a regex before checking if it's a directory in the dir-glob module?

It's hard to reliably do that. Let's skip that for now. I don't want to deal with the headache of false negatives.

sindresorhus:
Looks good! Glad we could finally include this in `globby` 😎

PR #48, opened 2017-10-18
Add `gitignore` option

marionebl:
* First introduced in sindresorhus/xo
* Known to be a performance bottleneck sindresorhus/xo#235
* Motivation: sindresorhus/xo#236


sindresorhus:
Yay! Glad to have this code consolidated in a central place.

Issue #50, opened 2017-10-22
Improve performance


sindresorhus:
This issue is dedicated to gathering ideas on how we can improve performance.

Some things that might help is to do filtering in the correct order to ensure minimal work is done, and cache and memoize as much as possible. The most important thing is that we skip ignored directories like `node_modules` as early as possible. It's the source of a *lot* of performance problems.

Some more ideas in #44 and mrmlnc/fast-glob#15

zachleat:
FWIW, the biggest performance bottleneck in my testing on [eleventy](https://github.com/11ty/eleventy/) (for whatever reason) was the `gitignore: true` option. Removing this and just looping over `.gitignore` myself and adding the entries as `!` lines was much faster, for whatever reason.

(Love globby—thanks!)

yaodingyd:
About filtering out `.gitignore` patterns before globbing, is translating gitignore rules and adding them to fast-glob ignore the right direction to go?

sindresorhus:
> About filtering out `.gitignore` patterns before globbing, is translating gitignore rules and adding them to fast-glob ignore the right direction to go?

I think so, yes.

sindresorhus:
> I'm adding this here as it's somewhat related: right now just requiring `globby` on my system often takes anywhere between 100ms and 250ms, and just installing it on a new directory downloads more than 4MB of stuff.

Yeah, that is quite unfortunate. It's because of `micromatch`, a dependency of `fast-glob`. See: https://github.com/mrmlnc/fast-glob/issues/154

<img width="924" alt="screen shot 2019-02-15 at 18 41 10" src="https://user-images.githubusercontent.com/170270/52854544-4d452580-3151-11e9-89d5-d5f8a62956ff.png">

// @mrmlnc

dwelle:
I concur with @zachleat that the `gitignore: true` has a huge perf cost for whatever reason.

On our codebase:

1. using `gitignore: true`:

    ~8sec
	
    ```js
    await globby([ `**` ], { gitignore: true });
    ```

2. custom:

    ~100ms

    ```js
    const ignored = fs.readFileSync(".gitignore", { encoding: "utf8" })
        .split("\n")
        .filter( line => !/^\s*$/.test(line) && !/^\s*#/.test(line))
        .map( line => "!" + line.trim().replace(/^\/+|\/+$/g, ""));

    const files = await globby([
        `**`,
        `!.git`,
        ...ignored
    ]);
    ```

~4800% difference is a lot...

[...]

PR #51, opened 2017-10-30
Add `fast-glob` to benchmarks

kevva:
In a step towards fixing #50, this might be useful to have. Currently it looks like `fast-glob` is faster than `globby`.


sindresorhus:
Speed is cool, but correctness is also important. `node-glob` is used in production considerably more than `fast-glob`. Let's reconsider it when https://github.com/mrmlnc/fast-glob/issues/15 is done.

PR #59, opened 2017-12-01
Add gitignoreName option

cheungpat:
This new option allows non-git tools to override the file name
of the .gitignore file when applying ignore patterns to globbed files.

The default is still `.gitignore` so the new option is compatible with
existing behavior.

### DIFF
--- a/fixtures/otherignore/.otherignore
+++ b/fixtures/otherignore/.otherignore
@@ -1,0 +1,2 @@
+foo.js
+!bar.js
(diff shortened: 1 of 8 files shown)


sindresorhus:
What do you need this for? We don’t add new options lightly.

cheungpat:
@sindresorhus We are making a [developer tool](https://github.com/SkygearIO/skycli) for deploying code to server and we want to ignore files using a similar mechanism as `.gitignore`. We cannot use the `.gitignore` file because the ignore patterns for deploy and for source control are different.

So we need an option to specify which file to use as ignore patterns. Without this, we need to fork globby or find similar libraries and combine them to do the same thing.

Yea, it is pretty bad to add more options. I hope you could still consider this.

PR #69, opened 2018-02-11
Replace node-glob by fast-glob

pvdlg:
Per [comment](https://github.com/sindresorhus/globby/pull/68#issuecomment-362805086)

### DIFF
--- a/gitignore.js
+++ b/gitignore.js
@@ -79,7 +79,7 @@ const normalizeOpts = opts => {
 module.exports = o => {
 	const opts = normalizeOpts(o);
 
-	return fastGlob('**/.gitignore', {ignore: opts.ignore.concat(DEFAULT_IGNORE), cwd: opts.cwd})
+	return fastGlob('**/.gitignore', {ignore: DEFAULT_IGNORE.concat(opts.ignore), cwd: opts.cwd})
 		.then(paths => Promise.all(paths.map(file => getFile(file, opts.cwd))))
 		.then(files => reduceIgnore(files))
 		.then(ignores => getIsIgnoredPredecate(ignores, opts.cwd));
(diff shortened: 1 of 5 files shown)


gitignore.js
```diff
@@ -1,12 +1,19 @@
 'use strict';
 const fs = require('fs');
 const path = require('path');
-const glob = require('glob');
+const fg = require('fast-glob');
```
kevva: Name it `fastGlob` so that it's easier to understand what it is right away.

kevva:
Fantastic :+1:

gitignore.js
```diff
@@ -72,7 +79,7 @@ const normalizeOpts = opts => {
 module.exports = o => {
 	const opts = normalizeOpts(o);
 
-	return globP('**/.gitignore', {ignore: opts.ignore, cwd: opts.cwd})
+	return fg('**/.gitignore', {ignore: opts.ignore.concat(DEFAULT_IGNORE), cwd: opts.cwd})
```
kevva: Do we want to be able to overwrite the default ones? I guess putting `DEFAULT_IGNORE` at the end would take precedence over the users?

index.js
```diff
@@ -1,11 +1,10 @@
 'use strict';
 const arrayUnion = require('array-union');
 const glob = require('glob');
-const pify = require('pify');
+const fg = require('fast-glob');
```
kevva: See comment above.

pvdlg:
@kevva thanks for the review. I made the changes you suggested.

kevva:
Also, I think [these](https://github.com/pvdlg/globby/blob/100a0906f6bf66830c07c60bd1aa100dd8b4f51a/index.js#L25-L28) could be removed since `fast-glob` doesn't seem to support them.

sindresorhus:
@pvdlg Yay! A new `globby` version is out. Would you be able to update `globby` in XO?

pvdlg:
yep, I'm on it!

Issue #70, opened 2018-02-13
Directory Expansion 8.X on Windows


calebboyd:
I see this passed in CI so it must just be a windows issue only

Running at 5d3d4de

![image](https://user-images.githubusercontent.com/5818726/36136613-062e133a-1057-11e8-98f6-8d24da1eb48f.png)

(expansion is working fine in latest 7.X (pending #63))

mrmlnc:
This is due to a stupid mistake in the `fast-glob` package. I forgot about converting the `\` symbol to UNIX style (`/`) in the patterns.

So, will be fixed today. Sorry for this situation.

mrmlnc:
@sindresorhus, may be you want [set-up](https://github.com/mrmlnc/globby/commit/84bd95f46c761b595b9f082f4534d2ab1f8fdcf9) [AppVeyor](https://www.appveyor.com) for globby?

sindresorhus:
I don't really want 😝, as AppVeyor is annoying to setup and use, and it's flaky, but yes, I'll do it.

mrmlnc:
Fixed by `fast-glob@2.0.4`.

PR #73, opened 2018-02-13
add .appveyor.yml

calebboyd:
re: https://github.com/sindresorhus/globby/issues/70#issuecomment-365316778

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,20 @@
+environment:
+  matrix:
+    - nodejs_version: '8'
+    - nodejs_version: '6'
+    - nodejs_version: '4'
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - npm install --global npm@latest
+  - npm install
+
+test_script:
+  - node --version
+  - npm --version
+  - npm test
+
+matrix:
+  fast_finish: true
+
+build: off
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,3 +10,4 @@ insert_final_newline = true
 [*.yml]
 indent_style = space
 indent_size = 2
+quote_style = single


kevva:
Maybe install a newer version of `npm` to make it faster on older Node.js versions? And also setting `fast_finish: true` since AppVeyor takes forever to run.

.appveyor.yml
```diff
@@ -0,0 +1,18 @@
+environment:
+  fast_finish: true
```
kevva: I think this should be under `matrix` (not this one under environment, just add it to the bottom), e.g:

```yml
matrix:
  fast_finish: true
```

.appveyor.yml
```diff
@@ -0,0 +1,18 @@
...
+  matrix:
+    - nodejs_version: "8"
+    - nodejs_version: "6"
+    - nodejs_version: "4"
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - npm install npm@latest -g
```
kevva: Nitpick, but better to be verbose with flags in code so it's easier to understand:

```
- npm install --global npm@latest
```

.appveyor.yml
```diff
@@ -0,0 +1,18 @@
+environment:
+  fast_finish: true
+  matrix:
+    - nodejs_version: "8"
```
kevva: Single quotes.

cah-calebboyd: I think there is an `.editorconfig` setting for that

kevva:
Looks good 👍

.appveyor.yml
```diff
@@ -0,0 +1,20 @@
+environment:
+  matrix:
+    - nodejs_version: '8'
+    - nodejs_version: '8'
```
evocateur: This line looks like a duplicate? Unless the intention was to test Node 8.x twice.

calebboyd: 😬 🤦‍♂️

Issue #53, opened 2017-11-16
Flip nodir to false if pattern ends with a trailing "/"


jdalton:
I noticed in globby 7 that `nodir` is `true` by default. It would be nice if it would auto-flip to `false` if the pattern has a trailing `"/"` which is used to indicate matching directories.

sindresorhus:
I really don't like that pattern. I know it's used by `cp` and some other UNIX tools, but it's very surprising for many users. I never postfix directory paths with `/`, for example.

It's better to be explicit about this. Especially because the patterns could come from a user, and the implementation might not even be able to handle directories. This option should rather explicitly be set by the developer using `globby`. Or better yet, just depend on the `expandDirectories` option and you don't need this at all. Most use-cases only need files, not directories.

jdalton:
I guess my usage is just going in a different direction than globby, v7 at least. For my use I don't want to expand folders and don't want "nodir" enabled. I'm cleaning up build folders and spare files where I delete whole directories at once before running unit tests or publishing. I based the trailing "/" suggestion off of the `nodir` [docs](https://github.com/isaacs/node-glob#options).

> `nodir` Do not match directories, only files. (Note: to match only directories, simply put a / at the end of the pattern.)

sindresorhus:
> I guess my usage is just going in a different direction than globby, v7 at least. For my use I don't want to expand folders and don't want "nodir" enabled.

I'm just optimizing the common case. You can get what you need with `{expandDirectories: false, nodir: false}`.

> I'm cleaning up build folders and spare files where I delete whole directories at once before running unit tests or publishing.

You can't remove directories with Node.js unless they're empty, so it will still need to expand the directories into files, remove the files, and then the directory, even if that happens at a different level.

jdalton:
I'm using fs-extra to remove dirs. Does nodir enabled auto disable expand?

sindresorhus:
`fs-extra` expands directories too: https://github.com/jprichardson/node-fs-extra/blob/f934357bb6a933fc47138e31492b1507bc026ff4/lib/remove/rimraf.js#L63 So it makes no difference if it's done here or in `fs-extra`.

---

> Does nodir enabled auto disable expand?

No, `nodir` enabled is the default. Meaning you will never get any directories. You can still set `nodir: false`, while `expandDirectories: true` if you also want the directory entry. For example, `foo/a`, `foo/b`, but also `foo`, if you do `globby('foo')`.

jdalton:
[...]

PR #76, opened 2018-02-20
docs: replace unsupported nodir option with onlyDirectories

chocolateboy:
fast-glob doesn't support glob's `nodir` option [1], but it supports an `onlyFiles` option which does the same thing, and an `onlyDirectories` option which does the inverse.

`onlyDirectories: true` is clearer in this case than `onlyFiles: false`, and `onlyDirectories` takes precedence over `onlyFiles` [2], so update the documentation to use that instead.

[1] https://github.com/mrmlnc/fast-glob#compatible-with-node-glob
[2] https://github.com/mrmlnc/fast-glob/pull/51

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -80,7 +80,7 @@ If set to `true`, `globby` will automatically glob directories for you. If you d
 })();
 ```
 
-Note that if you set this option to `false`, you won't get back matched directories unless you set `nodir: false`.
+Note that if you set this option to `false`, you won't get back matched directories unless you set [`onlyDirectories`](https://github.com/mrmlnc/fast-glob#onlydirectories) to `true`.
 
 ##### gitignore
 


sindresorhus:
> onlyDirectories: true is clearer in this case than onlyFiles: false, and onlyDirectories takes precedence over onlyFiles [2], so update the documentation to use that instead.

I don't think so. They do different things. `onlyFiles: false` returns files *and* directories. While `onlyDirectories: true` will *only* return directories.

chocolateboy:
You're right. Thanks for clarifying.

I'm going to close this as I can't find a simple way to explain how to wrangle the 3 (!) different options (`expandDirectories`, `onlyFiles`, and `onlyDirectories`) to do what [previously](https://github.com/sindresorhus/globby/issues/53) could be done entirely within the glob.

PR #78, opened 2018-03-16
Add typescript support.

DomiR:


sindresorhus:
I'm not using TypeScript personally, so I'm not interested or able to maintain type definitions for it. Would be better to submit to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) instead.

See: https://github.com/sindresorhus/ama/issues/439

sindresorhus:
I now accept TypeScript definitions if anyone looking is interested in submitting it here.

See: https://github.com/sindresorhus/ama/issues/439#issuecomment-414212229

Please follow this guide: https://github.com/sindresorhus/typescript-definition-style-guide

Issue #87, opened 2018-05-21
Expands directories is not working with cwd option


zeakd:
Hi. with different `cwd`, expands directories is not working.

```
└── a/
  ├── b
  └── c/
    ├── d
    └── e
```

``` js
// in directory 'a'

globby('c') // ['c/d', 'c/e']
globby('c', { cwd: process.cwd() }) // ['c/d', 'c/e']
globby('a', { cwd: path.resolve(process.cwd(), '..') }) //  []
globby('a/**/*', { cwd: path.resolve(process.cwd(), '..') }) // ['a/b', 'a/c/d', 'a/c/e']
```

thanks

thiagodp:
Same happening here, version `8.0.1` on Windows

pvdlg:
This is due to https://github.com/kevva/dir-glob/issues/7.
The fix has been done a while ago but it's not released on npm...

calebboyd:
@kevva can we bother you for an `npm publish`?

sindresorhus:
The `dir-glob` version is published. @pvdlg What needs to be done on our side?

pvdlg:
Yes we have to forward the `cwd` option (passed to `globby`) to `dir-glob`.

pvdlg:
@sindresorhus after looking in more detail we don't have to do anything other than making a new release of `globby`. The last released version is `8.0.2` which locked to `dir-glob` dependency to `2.0.0`. I don't why...

That has been reversed in commits following `8.0.2`, so if we make a new release of globby we should be good.

sindresorhus:
> The last released version is 8.0.2 which locked to dir-glob dependency to 2.0.0. I don't why...

See https://github.com/sindresorhus/globby/pull/99#issuecomment-452647305

sindresorhus:
I have done a new release.

PR #88, opened 2018-07-03
Fix `expandDirectories` and `ignores` option working together

### LINKED ISSUE #62: `expandDirectories` option should also apply to `ignore` patterns
[expanddirectories](https://github.com/sindresorhus/globby#expanddirectories) applied only on the patterns in the [patterns](https://github.com/sindresorhus/globby#patterns) options. Directories defined in the `ignore` options should be expanded too.

alexander-alvarez:
I came across https://github.com/sindresorhus/globby/issues/62 when encountering an issue with relative paths + ignores. I tried fixing the relative path issue, but I've concluded that seems to be in `fast-glob`, but it did seem to fix #62, so I'm PR-ing it and a failing test case for the relative path

Fixes #62

### DIFF
--- a/index.js
+++ b/index.js
@@ -60,6 +60,17 @@ const globDirs = (task, fn) => {
 
 const getPattern = (task, fn) => task.opts.expandDirectories ? globDirs(task, fn) : [task.pattern];
 
+const globToTask = task => glob => {
+	const opts = task.opts;
+	if (opts.ignore && Array.isArray(opts.ignore) && opts.expandDirectories) {
+		opts.ignore = dirGlob.sync(opts.ignore);
+	}
+	return {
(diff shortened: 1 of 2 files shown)


alexander-alvarez:
The appveyor seems unrelated

pvdlg:
The relative path issue is due to #87.
The problem has been fixed in `dir-glob` but never released. It seems the library is not maintained anymore: kevva/dir-glob#7

Maybe we could copy the code of `dir-glob` into `globby` so we could solve the problem completely?

sindresorhus:
@pvdlg I talked with the author of `dir-glob`. He lost his phone with 2FA codes a while ago. He's gonna try to get into his npm account and publish it.

PR #99, opened 2019-01-09
Fix compatibility with latest `dir-glob` release

doochik:
This bug fixes problem with latest dir-glob@2.2.0

Related to https://github.com/kevva/dir-glob/pull/9


doochik:
Thanks!

Issue #20, opened 2015-11-09
provide a `map` function for faster async behavior


jamestalmage:
```js
function handleMatch (individualMatch) {
  // return a promise
}

globby.map([...patterns], handleMatch).then(function (arrayOfPromiseResults)) {

});
```

essentially the equivalent of: 

```js
globby.map([...patterns]).then(function (arrayOfMatches) {
  return Promise.all(arrayOfMatches.map(handleMatch));
});
```

But it provides better async behavior by not requiring you wait for globby to be done before handling a match.

sindresorhus:
What do you need it for? Like, what kind of handling? Unless you're using globbing on a server the sync version is usually faster.

UltCombo:
This seems to be a dupe of #4, though you suggest a different API.

node-glob exposes an [EventEmitter API](https://github.com/isaacs/node-glob#class-globglob). Perhaps we could expose it as well.

For the time being, I'd suggest using [glob-stream](https://github.com/gulpjs/glob-stream) for that use case.

jamestalmage:
Didn't realize sync was faster.

I am scanning the entire Babel codebase multiple times. It's Huge.

Figured out my slowdown was only partially related to globby.

sindresorhus:
The only way to really know is to benchmark. You can almost never guess what's slowing down JS.

jamestalmage:
Well, in this particular case it was me being sloppy.

PR #104, opened 2019-01-14
Upgrade `ignore` dependency to version 5

### LINKED ISSUE #92: Update `ignore` package to latest version
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/globby/issues/92)

> There is a $20.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/globby/issues/92).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/globby/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

yaodingyd:
Fix #92


gitignore.js
```diff
@@ -44,7 +44,7 @@ const reduceIgnore = files => {
 	}, gitIgnore());
 };
 
-const getIsIgnoredPredecate = (ignores, cwd) => {
+const getIsIgnoredPredicate = (ignores, cwd) => {
```
sindresorhus: Why change this to something that is incorrect?

sindresorhus:
I haven't look into this closely, but I always get suspicious when you have to modify tests to get something working. Are you sure this is the correct way to solve it? I would have thought we needed to do some changes to `gitignore.js`, like in https://github.com/sindresorhus/globby/blob/71b9c58f86b4625b58ae241360016f4418dac288/gitignore.js#L48

sindresorhus:
https://github.com/kaelzhang/node-ignore#upgrade-4x---5x

yaodingyd:
The 5x breaking change is that for every path it uses `const REGEX_TEST_INVALID_PATH = /^\.*\/|^\.+$/` this regex to check invalid path and throws errors. So in the test case for "foo.js" after `path.relative` it returns `../../foo.js` which throws errors. I want to confirm if globby is used, all the file paths from fastglob would be relative to current cwd and we would not have this issue, and all I need to do is take care of the test file? 

Or we do need to consider this kind of path so in line 48 we need to check if path matches that regex?

sindresorhus:
> Or we do need to consider this kind of path so in line 48 we need to check if path matches that regex?

I don't know. Part of the task of fixing #92 is figuring that out.

sindresorhus:
@yaodingyd Did you figure it out?

yaodingyd:
I'm checking the input path: if it matches a relative path starting with `./` then I get relative path to `cwd`; otherwise I would pass it directly to `ignore`. I think my assumption makes sense: usually when we use globby, we would use it under `cwd` and any files path looks like the test ones, they should be under `cwd`, which is the results of `path.relative`, so that we can directly use them.

sindresorhus:
Closing in favor of #20

PR #107, opened 2019-01-24
Respect gitignore files from parent directories

### LINKED ISSUE #86: Respect `gitignore` from above
[Issuehunt badges](https://oss.issuehunt.io/r/sindresorhus/globby/issues/86)

Currently only gitignores from cwd and down are taken into account for gitignore filtering. What about those who are higher in a hierarchy?

Use case: monorepo repositories (like babel for example), where gitignore is placed in root and individual packages do not have their own.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [issuehunt issuehunt](https://oss.issuehunt.io/u/issuehunt) ($80.00)

### Submitted pull Requests
- [#107 include gitignore files from parent dirs](https://oss.issuehunt.io/r/sindresorhus/globby/pull/107)
- [#781 Allow `throwHttpErrors` option to be a function](https://oss.issuehunt.io/r/sindresorhus/ky/pull/781)
---

### Tips

- Checkout the [Issuehunt explorer](https://oss.issuehunt.io/r/sindresorhus/globby/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://oss.issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

dflupu:
Fixes #86

### DIFF
--- a/gitignore.js
+++ b/gitignore.js
@@ -58,8 +58,8 @@ const ensureAbsolutePathForCwd = (cwd, p) => {
 };
 
 const getIsIgnoredPredecate = (ignores, gitRoot, cwd) => {
-	return p => {
-		const pathWithCwd = ensureAbsolutePathForCwd(cwd, p);
+	return givenPath => {
+		const pathWithCwd = ensureAbsolutePathForCwd(cwd, givenPath);
 		const pathRelativeToGitRoot = path.relative(gitRoot, pathWithCwd);
 		return ignores.ignores(slash(pathRelativeToGitRoot));
 	};
@@ -98,8 +98,8 @@ module.exports = async options => {
 	options = normalizeOptions(options);
 	const {cwd} = options;
 
-	const gitDir = await findUp('.git', {cwd});
-	const gitRoot = gitDir ? path.dirname(gitDir) : '';
+	const gitDirectory = await findUp('.git', {cwd});
+	const gitRoot = gitDirectory ? path.dirname(gitDirectory) : '';
 
 	const gitIgnoreFilePaths = await Promise.all([
 		findUpAll('.gitignore', {
@@ -114,6 +114,7 @@ module.exports = async options => {
 	]);
 
 	const gitIgnoreFileContents = await Promise.all(
+		// TODO: Use .flat() once Node.js 12 is targetted
 		[].concat(...gitIgnoreFilePaths)
 			.map(p => path.relative(gitRoot, p))
 			.map(file => getFile(file, gitRoot))
@@ -127,8 +128,8 @@ module.exports.sync = options => {
 	options = normalizeOptions(options);
 	const {cwd} = options;
 
-	const gitDir = findUp.sync('.git', {cwd});
-	const gitRoot = gitDir ? path.dirname(gitDir) : '';
+	const gitDirectory = findUp.sync('.git', {cwd});
+	const gitRoot = gitDirectory ? path.dirname(gitDirectory) : '';
 
 	const gitIgnoreFilePaths = [
 		...findUpAll.sync('.gitignore', {
--- a/gitignore.js
+++ b/gitignore.js
@@ -5,6 +5,8 @@ const path = require('path');
 const fastGlob = require('fast-glob');
 const gitIgnore = require('ignore');
 const slash = require('slash');
+const findUp = require('find-up');
+const findUpAll = require('find-up-all');
 
 const DEFAULT_IGNORE = [
 	'**/node_modules/**',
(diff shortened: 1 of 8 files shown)


wavebeem:
I just came here to work on this issue and here's a PR already! This would really help out my jsx-info project.

sindresorhus:
@dflupu Hey, sorry. I totally missed this PR. If you're still interested, would you be able to fix the merge conflict and add some more tests?

dflupu:
Sure, I'll have it fixed up in the next few days.

dflupu:
@sindresorhus I've updated the PR. Let me know what you think.

gitignore.js
```diff
@@ -55,8 +57,12 @@ const ensureAbsolutePathForCwd = (cwd, p) => {
 	return path.join(cwd, p);
 };
 
-const getIsIgnoredPredecate = (ignores, cwd) => {
-	return p => ignores.ignores(slash(path.relative(cwd, ensureAbsolutePathForCwd(cwd, p))));
+const getIsIgnoredPredecate = (ignores, gitRoot, cwd) => {
+	return p => {
```
sindresorhus: Use a descriptive variable name.

gitignore.js
```diff
@@ -90,28 +96,55 @@ const normalizeOptions = ({
...
-		cwd: options.cwd
-	});
-
-	const files = await Promise.all(paths.map(file => getFile(file, options.cwd)));
-	const ignores = reduceIgnore(files);
-
-	return getIsIgnoredPredecate(ignores, options.cwd);
+	const {cwd} = options;
+
+	const gitDir = await findUp('.git', {cwd});
```
sindresorhus:
```suggestion
	const gitDirectory = await findUp('.git', {cwd});
```

gitignore.js
```diff
@@ -90,28 +96,55 @@ const normalizeOptions = ({
...
+		}),
+		fastGlob('**/.gitignore', {
+			ignore: DEFAULT_IGNORE.concat(options.ignore),
+			absolute: true,
+			cwd
+		})
+	]);
+
+	const gitIgnoreFileContents = await Promise.all(
+		[].concat(...gitIgnoreFilePaths)
```
sindresorhus: Can you add a TODO comment about using flatMap here when we target Node.js 12?

sindresorhus: I assume we do `[].concat(...gitIgnoreFilePaths)` because it's nested arrays?

sindresorhus:
The new behavior needs to be documented in the readme and index.d.ts.

And it would be good with some integrations tests that uses `globby()` and not `gitignore()`.

sindresorhus:
I'm also concerned that this will further worsen the performance, see: https://github.com/sindresorhus/globby/issues/50#issuecomment-594106877

dflupu:
> I'm also concerned that this will further worsen the performance, see: [#50 (comment)](https://github.com/sindresorhus/globby/issues/50#issuecomment-594106877)

I've done a bit of testing, and that may happen in some circumstances. Globbing small directory trees (eg <5 directories, <100 files) using the sync function may be up to 3x slower. It does however become less measurable when globbing larger directory trees.

The async function is pretty much unaffected when it comes to performance - unless there is a huge gitignore file in a parent directory.

Note that the main reason as to why the gitignore option is slow in the first place seems to be the globbing for gitignore files in subdirectories. Maybe that could be avoided in some way? How about checking for gitignore files only in directories and parent directories of files that were matched by the users' glob expression?

sindresorhus:
> Maybe that could be avoided in some way? How about checking for gitignore files only in directories and parent directories of files that were matched by the users' glob expression? Some kind of callback from fast-glob whenever it wants to descend into a directory that tells it whether or not it should do that could also work.

Yes, I think that would be smart.

sindresorhus:
> Some kind of callback from fast-glob whenever it wants to descend into a directory that tells it whether or not it should do that could also work.

Could you open an issue on `fast-glob` about this use-case? Maybe there are other things `fast-glob` could do to make this easier.

sindresorhus:
@dflupu Bump

dflupu:
@sindresorhus I've opened an issue on `fast-glob` at some point but nothing came out of that. I've also benchmarked a few different ways of dealing with the gitignore stuff and there was no clear winner. What worked better was tied to a lot of variables such as how many files were in the project, how complex the gitignore files were and also disk/cpu speed.

sindresorhus:
> I've opened an issue on fast-glob at some point but nothing came out of that.

Can you share a link to the issue?

sindresorhus:
How do we move forward with this PR?

dflupu:
Here's the [fast-glob issue](https://github.com/mrmlnc/fast-glob/issues/265). I'll mention that I've tried using `@nodelib/fs.walk` as the author suggested, and I've also partially implemented the functionality requested in the issue. Neither was an all-around clear winner in terms of performance (I don't remember the exact details as it's been a while).

I see a few options for this PR:
1) Include it under an optional flag
2) Keep it around until the overall performance of the gitignore option is somehow improved
3) Close it. Maybe someone else will pick up this up in the future if there's no open PR, and they'll also manage to somehow improve the performance of the gitignore option

sindresorhus:
I'm going to close as it doesn't look like something will happen with this anytime soon. 

https://github.com/sindresorhus/globby/issues/86#issuecomment-792259899

Issue #105, opened 2019-01-15
v9 not throwing ENOTDIR when cwd is a file


chrisblossom:
This might be expected, but when using ``'.'`` as the pattern and specifying `cwd` as a file, an error is no longer thrown in `v9` like it was in `v8`.

passing: [chrisblossom/globby/v8-is-file/test.js#L199-L221](https://github.com/chrisblossom/globby/blob/v8-is-file/test.js#L199-L221)
failing: [chrisblossom/globby/v9-is-file/test.js#L218-L241](https://github.com/chrisblossom/globby/blob/v9-is-file/test.js#L218-L241)

sindresorhus:
Would you be able to submit those tests as [failing tests](https://github.com/avajs/ava/blob/master/docs/01-writing-tests.md#failing-tests)? That would help getting this fixed faster.

IssuehuntBot:
@issuehunt has funded $20.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/21069494/issues/105)

medusalix:
This is caused by the update to [`dir-glob`](https://github.com/kevva/dir-glob) version 2.2.1. The commit [`Add cwd option`](https://github.com/kevva/dir-glob/commit/7230251a51181214f9cb8b2ae4601b96cbe693b4) added the cwd option to `dir-glob`, which changed the way it treats the filepath passed to it:

**2.0.0**

```js
const getPath = filepath => filepath[0] === '!' ? filepath.slice(1) : filepath;
```

**2.2.1**

```js
const getPath = (filepath, cwd) => path.join(cwd, filepath[0] === '!' ? filepath.slice(1) : filepath);
```

@sindresorhus What should we do? Remove the failing tests?

sindresorhus:
> What should we do?

Add a check to ensure `cwd` is a directory.

IssuehuntBot:
@sindresorhus has rewarded $18.00 to @medusalix. [See it on IssueHunt](https://issuehunt.io/repos/21069494/issues/105)

- :moneybag: Total deposit: $20.00
- :tada: Repository reward(0%): $0.00
- :wrench: Service fee(10%): $2.00

PR #108, opened 2019-01-29
Add failing test for #105

chrisblossom:
See https://github.com/sindresorhus/globby/issues/105

### DIFF
--- a/test.js
+++ b/test.js
@@ -214,3 +214,17 @@ test('respects gitignore option false - sync', t => {
 	const actual = m.sync('*', {gitignore: false, onlyFiles: false});
 	t.true(actual.indexOf('node_modules') > -1);
 });
+
+// https://github.com/sindresorhus/globby/issues/105
+test.failing('throws ENOTDIR when specifying a file as cwd - async', async t => {
+	const isFile = path.resolve('fixtures/gitignore/bar.js');
+	await t.throwsAsync(m('.', {cwd: isFile}), {code: 'ENOTDIR'});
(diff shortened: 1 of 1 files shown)


test.js
```diff
@@ -214,3 +214,29 @@ test('respects gitignore option false - sync', t => {
...
+test.failing('throws ENOTDIR when specifying a file as cwd - async', async t => {
+	t.plan(1);
+
+	const isFile = path.resolve('fixtures/gitignore/bar.js');
+
+	try {
+		await m('.', {cwd: isFile});
+	} catch (error) {
+		t.is('ENOTDIR', error.code);
+	}
```
sindresorhus: Use `t.throwsAsync` instead of try/catch. Then you can drop the `t.plan()`.

test.js
```diff
@@ -214,3 +214,29 @@ test('respects gitignore option false - sync', t => {
...
+test.failing('throws ENOTDIR when specifying a file as cwd - sync', t => {
+	t.plan(1);
+
+	const isFile = path.resolve('fixtures/gitignore/bar.js');
+
+	try {
+		m.sync('.', {cwd: isFile});
+	} catch (error) {
+		t.is('ENOTDIR', error.code);
+	}
```
sindresorhus: Use `t.throws` instead of try/catch. Then you can drop the `t.plan()`.

test.js
chrisblossom: Done.

test.js
chrisblossom: Done.

sindresorhus:
Thanks :)

PR #112, opened 2019-03-16
Upgrade `ignore` dependency

### LINKED ISSUE #92: Update `ignore` package to latest version

futpib:
fixes #92


sindresorhus:
Can you describe the changes? Also see https://github.com/sindresorhus/globby/pull/104 for a previous attempt that I don't think is correct.

futpib:
After upgrade you get invalid path errors you showed in https://github.com/sindresorhus/globby/issues/92#issuecomment-435317320. The [`ignore` upgrade guide](https://github.com/kaelzhang/node-ignore#upgrade-4x---5x) says they made paths starting with `./` or `../` invalid. I saw that these were introduced by `path.relative` call in `ignores.ignores(slash(path.relative(cwd, p)))` and changed it to `ignores.ignores(slash(p))`. At this point I trusted green tests and submitted the PR.

Now, after a closer look, I see that this almost made sense except for a call like `globby.sync('../foo/*', { gitignore: true }` and a `.gitignore` file in cwd, which is not covered by the tests, which would result in `ignores` call with a path starting with `../`.

I might come back to this with a better PR.

PR #125, opened 2019-06-21
Don't throw when specifying a non-existing `cwd` directory

popomore:


index.js
```diff
@@ -19,7 +19,7 @@ const assertPatternsInput = patterns => {
 };
 
 const checkCwdOption = options => {
-	if (options && options.cwd && !fs.statSync(options.cwd).isDirectory()) {
+	if (options && options.cwd && fs.existsSync(options.cwd) && !fs.statSync(options.cwd).isDirectory()) {
```
sindresorhus: This is wasteful. You're doing two fs calls. It could also cause a race issue where the file system might change between the calls. Instead, you can just try/catch `fs.statSync.`

test.js
```diff
@@ -362,3 +362,13 @@ test('throws when specifying a file as cwd - stream', t => {
 		globby.stream('*', {cwd: isFile});
 	}, 'The `cwd` option must be a path to a directory');
 });
+
+test('ignore when specifying a noexist directory as cwd - async', async t => {
+	const actual = await globby('.', {cwd: '/unknown'});
+	t.true(actual.length === 0);
```
sindresorhus: Use `t.is`

test.js
```diff
@@ -362,3 +362,13 @@ test('throws when specifying a file as cwd - stream', t => {
 		globby.stream('*', {cwd: isFile});
 	}, 'The `cwd` option must be a path to a directory');
 });
+
+test('ignore when specifying a noexist directory as cwd - async', async t => {
```
sindresorhus:
```suggestion
test('don't throw when specifying a non-existing cwd directory - async', async t => {
```

sindresorhus:
The PR needs a better title.

popomore:
@sindresorhus PR is updated.

index.js
```diff
@@ -18,8 +18,19 @@ const assertPatternsInput = patterns => {
...
-const checkCwdOption = options => {
-	if (options && options.cwd && !fs.statSync(options.cwd).isDirectory()) {
+const checkCwdOption = (options = {}) => {
+	if (!options.cwd) {
+		return;
+	}
+
+	let stat;
+	try {
+		stat = fs.statSync(options.cwd);
```
kevva: Do we want to do this synchronously?

sindresorhus: We probably want a separate async version. I might have been too lazy to do that initially or just forgot. PR welcome :)

kevva: Yeah, I think we'll need it since we use `globby` in a lot of modules and this might result in a decrease in performance.

PR #132, opened 2019-07-17
fix: readonly of typescript definition

naughtLdy:
fix TypeScript Definition.


sindresorhus:
There’s nothing to fix. Just upgrade TypeScript.

Issue #133, opened 2019-07-17
Path *** is not in cwd ***


fjc0k:
Just running this code:

```js
const paths = await globby(
 '../src/*.ts',
  {
    dot: true,
    onlyFiles: false,
    gitignore: true,
    cwd: __dirname,
    absolute: true,
  },
)
```

Then you will get an error:

```bash
Path *** is not in cwd ***
```

But if you set `gitignore` to `false`, all works well.

After a bit of investigation, most likely it is caused by this PR: https://github.com/sindresorhus/globby/pull/120

sindresorhus:
// @jamiekyle-eb

jaydenseric:
I'm seeing this error too, when trying to use an glob containing an absolute `os.tmpdir()` path with `{ gitignore: true }`.

Absolute temp dir paths are very useful in tests.

rwjblue:
https://github.com/sindresorhus/globby/pull/137 fixes at least some of the scenarios where this error is being thrown, though it isn't clear if it is all of them.

sindresorhus:
@fjc0k Can you try out https://github.com/sindresorhus/globby/releases/tag/v10.0.2 and let us know if that fixes your issue?

fjc0k:
@sindresorhus  No, always the same error for each version:

`10.0.0`, `10.0.1`, `10.0.2`, `11.0.0`:

![image](https://user-images.githubusercontent.com/13151189/71820008-ad75ea80-30c8-11ea-9815-4c1783b23839.png)

nzakas:
I'm also seeing a version of this issue. In my case, it seems to be caused by my being on Windows, so the absolute path being checked and the cwd are not in the same format. That means using `startsWith()` to check the two paths doesn't work (https://github.com/sindresorhus/globby/blob/master/gitignore.js#L48) (and I would expect would cause other errors).

The error I'm getting:

```
Path C:/Users/Nicholas/Documents/projects/nitpik/cli/nitpik.config.js is not in cwd C:\Users\Nichol
as\Documents\projects\nitpik\cli
```

nzakas:
I opened a PR to at least solve the issue in Windows I was seeing:
https://github.com/sindresorhus/globby/pull/143

I'm not sure if it helps with other use cases, too.

navoneel07:
> I'm seeing this error too, when trying to use an glob containing an absolute `os.tmpdir()` path with `{ gitignore: true }`.
> 
[...]

PR #137, opened 2019-10-16
Fix using `gitignore` and `absolute` options at the same time on Windows

rwjblue:
When `globby` is invoked with `globby('*', { absolute: true, gitignore: true })` on Windows systems, the assertion added in 4af038edf82f6be9fe71f003e5db53dae879e687 **always** triggers. This is because when we default the `cwd` in `normalizeOptions`, we didn't convert from `C:\\` style paths to forward slashes.

A simple work around for users hitting this issue, is to provide a `cwd` option that has already been converted:

```js
globby('*', { absolute: true, gitignore: true, cwd: slash(process.cwd()) })
```
---

The test added here was failing on Windows, example output prior to the fixes:

```
  test » gitignore option with absolute option
  C:\Users\travis\build\rwjblue\globby\gitignore.js:52
   51:
   52:     throw new Error(`Path ${p} is not in cwd ${cwd}`);
   53:   }
  Rejected promise returned by test. Reason:
  Error {
    message: 'Path C:/Users/travis/build/rwjblue/globby/a.tmp is not in cwd C:\\Users\\travis\\build\\rwjblue\\globby',
  }
```

Related to #133 (though it isn't clear that this fixes **all** of the scenarios that #133 is exploring).

### DIFF
--- a/gitignore.js
+++ b/gitignore.js
@@ -83,7 +83,7 @@ const getFileSync = (file, cwd) => {
 
 const normalizeOptions = ({
 	ignore = [],
-	cwd = process.cwd()
+	cwd = slash(process.cwd())
 } = {}) => {
 	return {ignore, cwd};
 };
--- a/test.js
+++ b/test.js
@@ -298,6 +298,11 @@ test('gitignore option with stats option', async t => {
 	t.false(actual.includes('node_modules'));
 });
 
+test('gitignore option with absolute option', async t => {
+	const result = await globby('*', {gitignore: true, absolute: true});
+	t.false(result.includes('node_modules'));
+});
+
 test('respects gitignore option false - stream', async t => {
 	const actual = await getStream.array(globby.stream('*', {gitignore: false, onlyFiles: false}));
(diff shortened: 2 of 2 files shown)


jamiekyle-eb:
I found a library [path-is-inside](https://github.com/domenic/path-is-inside) that is supposed to do a better job than the `filePath.startsWith(cwd)` check I wrote on https://github.com/sindresorhus/globby/blob/master/gitignore.js#L48

rwjblue:
Ya, looks like that one deals with case sensitivity by platform and normalizing trailing slashes, but still effectively is doing `.startsWith` (it uses `.lastIndexOf(needle, 0) === 0`). I _think_ it would fail in the same way described in my description above (because it would not convert the backslashes in `process.cwd()` to forward slashes which is what fast-glob is giving us).

sindresorhus:
I have a similar package we could use: https://github.com/sindresorhus/is-path-inside It fixes a couple of shortcomings of `path-is-inside`.

PR #154, opened 2020-10-02
Upgrade dev dependencies

iamnapo:
- Updated devDeps to latest versions.
- Refactored to match latest xo code-style.
- Refactored to fix breaking changes of ava v3.
- Also test against Node 14. (Only a tiny change in `fs.writeFileSync`).

### DIFF
--- a/index.js
+++ b/index.js
@@ -56,7 +56,7 @@ const generateGlobTasks = (patterns, taskOptions) => {
 
 		const ignore = patterns
 			.slice(index)
-			.filter(x => isNegative(x))
+			.filter(pattern => isNegative(pattern))
 			.map(pattern => pattern.slice(1));
 
 		const options = {
--- a/index.js
+++ b/index.js
@@ -138,17 +138,20 @@ module.exports = async (patterns, options) => {
 module.exports.sync = (patterns, options) => {
 	const globTasks = generateGlobTasks(patterns, options);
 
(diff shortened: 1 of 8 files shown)


sindresorhus:
Sure, go ahead with the refactor.

index.js
```diff
@@ -56,7 +56,7 @@ const generateGlobTasks = (patterns, taskOptions) => {
 
 		const ignore = patterns
 			.slice(index)
-			.filter(isNegative)
+			.filter(x => isNegative(x))
```
sindresorhus, line 59:
```suggestion
			.filter(pattern => isNegative(pattern))
```

sindresorhus:
Thanks :)

index.js
```diff
@@ -25,7 +25,7 @@ const checkCwdOption = (options = {}) => {
 	let stat;
 	try {
 		stat = fs.statSync(options.cwd);
-	} catch (_) {
+	} catch {
```
andi1984, line 28: @sindresorhus I think this issue slipped through with the removed brackets which fails now in the latest 3.0.2 version. Should I make a PR for fixing it?

andi1984 (commented):
IMHO there was a bug slipping through in a try-catch block which breaks now the latest release version 3.0.2.

sindresorhus:
https://github.com/sindresorhus/globby/pull/167#issuecomment-756388829

Issue #155, opened 2020-11-08
globby is b0rked on Windows: .sync nor .async deliver /any/ result.


GerHobbelt:
Related to #152, but `.sync` is broken as well: doesn't matter what you try, globby won't produce any results.

Sample code:

```
var testset = globby.sync(__dirname + '/specs/*.jisonlex');
console.error({testset, dir: path.normalize(__dirname + '/specs/')});
```

produces this:

```
{
  testset: [],
  dir: 'W:\\Projects\\sites\\library.visyond.gov\\80\\lib\\tooling\\jison\\packages\\lex-parser\\tests\\specs\\'
}
```

while that directory is filled with tens of *.jisonlex files.

Ditto for the original code:

```
  var testset = globby.sync([
    __dirname + '/specs/*.jison',
    __dirname + '/specs/*.lex',
    __dirname + '/specs/*.jisonlex',
    __dirname + '/specs/*.json5',
    '!'+ __dirname + '/specs/*-ref.json5',
    __dirname + '/specs/*.js',
    __dirname + '/lex/*.jisonlex',
  ]);
console.error({testset, dir: __dirname});
```

papb:
What NodeJS version? Did you try other versions?

GerHobbelt:
```
$ node --version
v12.18.4
```

No, haven't tried other versions, but have seen this issue before. Then I switched tactics as I was in a hurry to get things done, so I don't expect this to be a node version specific thing.

GerHobbelt:
Side Note: globby use like this (i.e. without **directories** in the search argument) works as expected:

```js
globby([ 'ebnf-parser*.js' ])
```

GerHobbelt:
New info

## Worky 😥 <sup>(phew!)</sup>

This works on Windows:

```
process.chdir(__dirname);
var testset = globby.sync(['./specs/*.jisonlex', './lex/*.jisonlex']);
```

As long as you 
- DO NOT pass Windows path separators ('\\') as delivered by `path.normalize()` et al 💢 
- DO NOT pass Windows drive / network share identifiers (e.g. 'W:' in 'W:\\' which would be a absolute path in Windows) into globby **anywhere**, you'll be fine.

## Not Worky 1

Ergo: doing something like

```
let basepath = path.normalize(path.join(__dirname, ......)).replace(/\\\\/g, '/'); // UNIXify path
// ^^^ INSUFFICIENT, as the Windows drive letter plus colon, e.g. 'W:/' will make it through
[...]

PR #156, opened 2020-11-22
fix: quick and dirty absolute windows support

phawxby:
Quick and dirty fix for #155


sindresorhus:
https://github.com/sindresorhus/globby/issues/155#issuecomment-732961466

sindresorhus:
The correct fix is something like https://github.com/mrmlnc/fast-glob/issues/240

PR #163, opened 2021-01-06
Update index.js

vladimir-ivanov:
removing the brackets is a breaking change in lower versions of nodejs


vladimir-ivanov:
please either accept this PR and re publish to npm or mark it as a breaking change in package.json instead as it breaks a lot of our projects which rely on rname-cli which loads ^11.0.0, thanks

PR #164, opened 2021-01-06
Update index.js

vladimir-ivanov:
fails lower versions of nodejs - e.g. we use node 8


sindresorhus:
Please don't spam with multiple PRs.

PR #165, opened 2021-01-06
Update index.js

vladimir-ivanov:
fails lower versions of nodejs - e.g. we use node 8

### DIFF
--- a/index.js
+++ b/index.js
@@ -25,7 +25,8 @@ const checkCwdOption = (options = {}) => {
 	let stat;
 	try {
 		stat = fs.statSync(options.cwd);
-	} catch {
+	} catch (error) {
+		console.warn(error);
 		return;
 	}
 


vladimir-ivanov:
please merge ASAP and do a new npm release if possible? breaks a lot of libraries for us, upgrading nodejs is not an option, or instead mark it as a breaking change?
https://github.com/jhotmann/node-rename-cli/blob/master/package.json#L30
breaks in our case
many thanks

sindresorhus:
This package targets Node.js 10: https://github.com/sindresorhus/globby/blob/4bb7f4a3ffb1a4ef23c86fbf70261110513e1fcc/package.json#L14

sindresorhus:
> instead mark it as a breaking change?

It was: https://github.com/sindresorhus/globby/releases/tag/v11.0.0

PR #171, opened 2021-03-21
Fix Windows compatibility for gitignore option

### LINKED ISSUE #153: ensureAbsolutePathForCwd method issue
Greetings,

I've faced a problem while performing a nuxt generation. I've fixed it for my case, but I felt like I should report this.

cwd is... C:/Documents/ExampleFolder/...
p is...     C:\Documents\ExampleFolder\...

I had to add this line of command to make it work.
cwd = cwd.toString().replace(/\\/gi,'/');

You might wanna check if cwd contains "\\" character, or you will most probably provide a better solution.

barak007:
Fixes #153

### DIFF
--- a/gitignore.js
+++ b/gitignore.js
@@ -48,7 +48,7 @@ const reduceIgnore = files => {
 const ensureAbsolutePathForCwd = (cwd, p) => {
 	cwd = slash(cwd);
 	if (path.isAbsolute(p)) {
-		if (p.startsWith(cwd)) {
+		if (slash(p).startsWith(cwd)) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
This needs a test.

barak007:
Sorry for the description-less PR. I was in a harry and forgot to write a description. 
This should fix a case that `ensureAbsolutePathForCwd` received a cwd Windows path and a path to check which is also Windows path.

The whole slashes approach is usually not that promising. but I couldn't change the entire codebase.

PR #172, opened 2021-03-25
Add `ignoreFiles` option

jridgewell:
This is very similar to the `gitignore` option, except it allows globing for any compatible ignore file (eg, `.prettierignore` or `.eslintignore`). AMP had a recent need for this with our `prettier` integration, because it doesn't respect its ignore file when using the API directly.

### DIFF
--- a/ignore.js
+++ b/ignore.js
@@ -2,7 +2,7 @@ import process from 'node:process';
 import fs from 'node:fs';
 import path from 'node:path';
 import fastGlob from 'fast-glob';
-import gitIgnore from 'ignore';
+import ignore from 'ignore';
 import slash from 'slash';
 import {toPath} from './utilities.js';
 
@@ -35,7 +35,7 @@ const parseGitIgnore = (content, options) => {
 };
 
 const reduceIgnore = files => {
-	const ignores = gitIgnore();
+	const ignores = ignore();
 	for (const file of files) {
 		ignores.add(parseGitIgnore(file.content, {
 			cwd: file.cwd,
--- a/ignore.js
+++ b/ignore.js
@@ -77,10 +77,10 @@ const normalizeOptions = (options = {}) => ({
 	cwd: toPath(options.cwd) || slash(process.cwd()),
 });
 
-export const isGitIgnored = async options => {
+export const isIgnored = async (pattern, options) => {
 	const {cwd} = normalizeOptions(options);
 
-	const paths = await fastGlob('**/.gitignore', {cwd, ...gitignoreGlobOptions});
+	const paths = await fastGlob(pattern, {cwd, ...gitignoreGlobOptions});
 
 	const files = await Promise.all(paths.map(file => getFile(file, cwd)));
 	const ignores = reduceIgnore(files);
@@ -88,10 +88,10 @@ export const isGitIgnored = async options => {
 	return getIsIgnoredPredicate(ignores, cwd);
 };
 
-export const isGitIgnoredSync = options => {
+export const isIgnoredSync = (pattern, options) => {
 	const {cwd} = normalizeOptions(options);
 
-	const paths = fastGlob.sync('**/.gitignore', {cwd, ...gitignoreGlobOptions});
+	const paths = fastGlob.sync(pattern, {cwd, ...gitignoreGlobOptions});
 
 	const files = paths.map(file => getFileSync(file, cwd));
 	const ignores = reduceIgnore(files);
--- a/index.d.ts
+++ b/index.d.ts
@@ -42,10 +42,19 @@ export interface Options extends FastGlobOptionsWithoutCwd {
 
 	/**
 	Respect ignore patterns in `.gitignore` files that apply to the globbed files.
+	This option takes precedence over the `ignoreFiles` option.
 
 	@default false
 	*/
 	readonly gitignore?: boolean;
+
+	/**
+	A glob pattern to look for ignore files, which are then used to ignore globbed files. This is a more generic form of the `gitignore` option, allowing you to find ignore files with a [compatible syntax](http://git-scm.com/docs/gitignore). For instance, this works with Babel's `.babelignore`, Prettier's `.prettierignore`, or ESLint's `.eslintignore` files.
+	This value is unused if `{ gitignore: true }`, since `gitignore` takes precedence.
+
+	@default undefined
+	*/
+	readonly ignoreFiles?: string;
 
 	/**
 	The current working directory in which to search.
--- a/index.d.ts
+++ b/index.d.ts
@@ -42,7 +42,7 @@ export interface Options extends FastGlobOptionsWithoutCwd {
 
 	/**
 	Respect ignore patterns in `.gitignore` files that apply to the globbed files.
-	This option takes precedence over the `ignoreFiles` option.
+	This option merges with `ignoreFiles` is both are specified.
 
 	@default false
 	*/
@@ -50,11 +50,11 @@ export interface Options extends FastGlobOptionsWithoutCwd {
 
 	/**
 	A glob pattern to look for ignore files, which are then used to ignore globbed files. This is a more generic form of the `gitignore` option, allowing you to find ignore files with a [compatible syntax](http://git-scm.com/docs/gitignore). For instance, this works with Babel's `.babelignore`, Prettier's `.prettierignore`, or ESLint's `.eslintignore` files.
-	This value is unused if `{ gitignore: true }`, since `gitignore` takes precedence.
+	This option merges with `.gitignore` if both are specified.
 
 	@default undefined
 	*/
-	readonly ignoreFiles?: string;
+	readonly ignoreFiles?: string | string[];
 
 	/**
 	The current working directory in which to search.
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ const assertPatternsInput = patterns => {
 };
 
 const toPatternsArray = patterns => {
-	patterns = [...new Set([patterns].flat())];
+	patterns = [...new Set([patterns].flat(Infinity))];
 	assertPatternsInput(patterns);
 	return patterns;
 };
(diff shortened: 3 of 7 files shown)


index.js
```diff
@@ -167,15 +180,20 @@ module.exports.stream = (patterns, options) => {
 	const filterStream = new FilterStream(p => !filter(p));
 	const uniqueStream = new UniqueStream();
 
-	return merge2(tasks.map(task => fastGlob.stream(task.pattern, task.options)))
+	return merge2(
+		tasks.map(task => fastGlob.stream(task.pattern, task.options))
+	)
```
sindresorhus, line 185: Don't do unrelated changes

index.d.ts
```diff
@@ -39,6 +39,13 @@ declare namespace globby {
 		@default false
 		*/
 		readonly gitignore?: boolean;
+
+		/**
+		Glob pattern to look for ignore files, which are then used to ignore globbed files.
```
sindresorhus, line 44: The definition of what "ignore files" are is not clear. The docs need to be expanded. There are also many unanswered questions: How does this relate to `gitignore` option. Does it work in addition to it, instead of it, etc.

You also need to add docs to the readme.

index.d.ts
```diff
@@ -39,6 +39,13 @@ declare namespace globby {
...
+
+		@default undefined
```
sindresorhus, line 46: This is not correct.

sindresorhus (changes requested):

index.js
jridgewell: Undone.

index.d.ts
jridgewell: Can you explain? If `ignoreFiles` isn't specified, we don't do anything with the value. Would you rather this say "empty string", or something else?

index.d.ts
jridgewell: Added a quick explanation here, and a larger explanation to the readme.

index.js
```diff
@@ -91,17 +91,25 @@ const globDirs = (task, fn) => {
...
 };
 
 const globToTask = task => glob => {
 	const {options} = task;
-	if (options.ignore && Array.isArray(options.ignore) && options.expandDirectories) {
+	if (
+		options.ignore &&
+		Array.isArray(options.ignore) &&
+		options.expandDirectories
+	) {
```
sindresorhus, line 112: Please don't do unrelated changes. Undo all unrelated changes.

sindresorhus:
You need to update package.json when renaming a file.

index.d.ts
```diff
@@ -39,6 +39,14 @@ declare namespace globby {
 		@default false
 		*/
 		readonly gitignore?: boolean;
+
+		/**
+		Glob pattern to look for ignore files, which are then used to ignore globbed files.
+		This value is unused if gitignore option is used, since gitignore takes precedence.
```
sindresorhus, line 45: This is ambiguous, setting `gitignore: false` is technically *it being used*. It should explicitly say that setting `gitignore: true` overrides this option.

readme.md
```diff
@@ -87,7 +87,14 @@ Note that if you set this option to `false`, you won't get back matched director
...
 
-Respect ignore patterns in `.gitignore` files that apply to the globbed files.
+Respect ignore patterns in `.gitignore` files that apply to the globbed files. This option takes precedence over the `ignoreFiles` option, and explicitly looks for files using a `**/.gitignore` glob.
+
+##### ignoreFiles
+
+Type: `string`\
+Default: `undefined`
+
+A glob pattern to look for ignore files, which are then used to ignore globbed files. This is a more generic form of the `gitignore` option, allowing you to find ignore files with a [compatible syntax](http://git-scm.com/docs/gitignore). For instance, this is works with Babel's `.babelignore`, Prettier's `.prettierignore`, or ESLint's `.eslintignore` files.
```
sindresorhus, line 97: Readme docs and index.d.ts docs should be in sync.

index.d.ts
sindresorhus: Sorry, I don't remember. I should have been clearer.

sindresorhus (changes requested):

readme.md
```diff
@@ -10,7 +10,7 @@ Based on [`fast-glob`](https://github.com/mrmlnc/fast-glob) but adds a bunch of
 - Multiple patterns
 - Negated patterns: `['foo*', '!foobar']`
 - Expands directories: `foo` → `foo/**/*`
-- Supports `.gitignore`
+- Supports `.gitignore`, or any similar ignore file system.
```
sindresorhus, line 13: This reads like "file system", not "ignore file  .. system". Use a different word than system.

index.js
jridgewell: Undone. It's the dang autoformatter.

index.d.ts
jridgewell: Done.

readme.md
jridgewell: Synced.

readme.md
jridgewell: Changed to "configuration", it was the best I could think of.

jridgewell (commented):
Addressed all comments, and made sure to undo any autoformatting (sorry about that).

weyert:
How can I help getting this PR merged into the project?

jridgewell:
I've rebased, this just requires a review.

index.d.ts
```diff
@@ -42,11 +42,20 @@ export interface Options extends FastGlobOptionsWithoutCwd {
...
 	*/
 	readonly gitignore?: boolean;
 
+	/**
+	A glob pattern to look for ignore files, which are then used to ignore globbed files. This is a more generic form of the `gitignore` option, allowing you to find ignore files with a [compatible syntax](http://git-scm.com/docs/gitignore). For instance, this works with Babel's `.babelignore`, Prettier's `.prettierignore`, or ESLint's `.eslintignore` files.
+	This value is unused if `{ gitignore: true }`, since `gitignore` takes precedence.
+
+	@default undefined
+	*/
+	readonly ignoreFiles?: string;
```
fisker, line 57: Why not take `string[] | sting` for easier use.

ignore.js
```diff
@@ -77,21 +77,21 @@ const normalizeOptions = (options = {}) => ({
...
 
 	return getIsIgnoredPredicate(ignores, cwd);
 };
 
-export const isGitIgnoredSync = options => {
+export const isIgnoredSync = (pattern, options) => {
 	const {cwd} = normalizeOptions(options);
 
-	const paths = fastGlob.sync('**/.gitignore', {cwd, ...gitignoreGlobOptions});
+	const paths = fastGlob.sync(pattern, {cwd, ...gitignoreGlobOptions});
```
fisker, line 94: Maybe we should add `dot: true` to the option, when pattern is `*ignore`, not sure, need test.

index.d.ts
```diff
@@ -42,11 +42,20 @@ export interface Options extends FastGlobOptionsWithoutCwd {
 
 	/**
 	Respect ignore patterns in `.gitignore` files that apply to the globbed files.
+	This option takes precedence over the `ignoreFiles` option.
```
fisker, line 45: This don't have to "takes precedence", we can concat `**/.gitignore` with `ignoreFiles`, so they can use together.

index.d.ts
jridgewell: Done.

index.d.ts
jridgewell: Done.

ignore.js
jridgewell: > Maybe we should add dot: true to the option, when pattern is `*ignore`

Sorry, I don't understand the suggestion.

> need test.

Added.

fisker: I mean, if user use `{ignoreFiles: '*ignore'}`, can this glob files starts with `.`? See https://github.com/mrmlnc/fast-glob#dot

index.js
```diff
@@ -14,7 +14,7 @@ const assertPatternsInput = patterns => {
 };
 
 const toPatternsArray = patterns => {
-	patterns = [...new Set([patterns].flat())];
+	patterns = [...new Set([patterns].flat(Infinity))];
```
fisker, line 17: Instead of flat infinity, why not `toPatternsArray(options.ignoreFiles)` then push `**/.gitignore`? Changing this will make `patterns` for `globby()` behave differently.

ignore.js
jridgewell: I personally don't find that necessary, and it makes the configuration a little complicated (what if the user didn't want to use dotfiles?). But I'll leave the decision to you.

index.js
jridgewell: Done.

ignore.js
fisker: I'm not a maintainer :)

index.js
```diff
@@ -52,13 +52,32 @@ const normalizeOptions = (options = {}) => {
...
 
-const getFilterSync = options => createFilterFunction(
-	options.gitignore && isGitIgnoredSync({cwd: options.cwd}),
-);
+	const patterns = toPatternsArray(options.ignoreFiles || []);
+	if (options.gitignore) {
+		patterns.push('**/.gitignore');
+	}
+
+	return toPatternsArray(patterns);
```
fisker, line 65:
```suggestion
	return patterns;
```

index.js
```diff
@@ -200,7 +219,6 @@ export const isDynamicPattern = normalizeArgumentsSync(
 export const generateGlobTasks = normalizeArguments(generateTasks);
 export const generateGlobTasksSync = normalizeArgumentsSync(generateTasksSync);
 
-export {
-	isGitIgnored,
-	isGitIgnoredSync,
-} from './gitignore.js';
+export {isIgnored, isIgnoredSync} from './ignore.js';
```
fisker, line 222: I'm not sure if we want to expose these, if we do, maybe we need a better name, `isIgnoredByIgnoreFiles`?

index.js
jridgewell: Ha.

index.d.ts
fisker: The content in readme.md file didn't fix.

jridgewell:
This has gotten another merge conflict. I'm not interested in doing more rounds. If you want to take this up, I release my changes as public domain.

fisker:
I can continue work on this.

Issue #139, opened 2020-01-03
type declarations do not account for fs.Stats[] return type


starpit:
fast-glob supports `{ stats: true }`, which yields `fs.Stats` objects as opposed to strings. it looks like `index.d.ts` does not account for this?

>>> Type 'string[]' is not assignable to type 'Stats[]'.
  Type 'string' is not assignable to type 'Stats'.

sindresorhus:
Probably just an oversight. PR welcome to fix this.

alex-page:
@starpit I had a go at fixing this in #173, let me know if you see any issues. I am new to to typescript.

PR #173, opened 2021-04-03
Add return type for Stats

alex-page:
This is my first time adding types to an open source package so there might be some mistakes here. I took a go at fixing  https://github.com/sindresorhus/globby/issues/139

This should fix up the use case of that should be returning `Stats` instead of `string`.
```
globby('*.tmp', {stats: true});
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -89,10 +89,13 @@ declare const globby: {
 	@param options - See the [`fast-glob` options](https://github.com/mrmlnc/fast-glob#options-3) in addition to the ones in this package.
 	@returns The matching paths.
 	*/
-	sync: (
+	sync: ((
 		patterns: string | readonly string[],
-		options?: {stats: true} & globby.GlobbyOptions
-	) => Stats[];
+		options: globby.GlobbyOptions & { stats: true }
+	) => Stats[]) & ((
+		patterns: string | readonly string[],
+		options?: globby.GlobbyOptions
+	) => string[]);
 
 	/**
 	Find files and directories using glob patterns.
@@ -168,10 +171,13 @@ declare const globby: {
 	})();
 	```
 	*/
-	(
+	((
 		patterns: string | readonly string[],
-		options?: {stats: true} & globby.GlobbyOptions
-	): Promise<Stats[]>;
+		options: globby.GlobbyOptions & { stats: true }
+	) => Promise<Stats[]>) & ((
+		patterns: string | readonly string[],
+		options?: globby.GlobbyOptions
+	) => Promise<string[]>);
 };
 
 export = globby;
--- a/index.d.ts
+++ b/index.d.ts
@@ -91,8 +91,8 @@ declare const globby: {
 	*/
 	sync: (
 		patterns: string | readonly string[],
-		options?: globby.GlobbyOptions
-	) => string[];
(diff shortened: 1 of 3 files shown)


sindresorhus:
It needs to return `Stats` only when `stats: true`.

alex-page:
@sindresorhus I am reading some documentation on this now. Is there a recommended approach or example repo you could nudge me in the right direction?

sindresorhus:
https://github.com/sindresorhus/query-string/blob/main/index.d.ts

alex-page:
I hit a wall. I have adjusted the test case and just need the `string[]` return for when `{stats: true}` does not exist. 

The declare const is making this a lot more tricky for me as I can't follow the typescript function overloading you mentioned. I am likely missing something.
https://github.com/sindresorhus/globby/blob/c1a3b3244ae992ba7e7e76f501a510ea0d9306df/index.d.ts#L81

Sorry for the extra noise.

sindresorhus:
Instead of replacing the methods, add overloads.

alex-page:
I spent a couple hours trying different implementations I can't figure it out how to add overloads to an already existing method 😔. I am going to keep moving as this isn't a blocking problem for me.

I tried a few different approaches:
```ts
declare function sync(
	patterns: string | readonly string[],
	options?: {stats: true} & globby.GlobbyOptions
): Stats[];

declare function sync(
	patterns: string | readonly string[],
	options?: globby.GlobbyOptions
) : string[] | Stats[];

declare const globby: {
	sync: typeof sync;
```

olup:
hey @alex-page, what about: 
```
	sync: ((
		patterns: string | readonly string[],
		options?: {stats : true} & globby.GlobbyOptions
	) => Stats[]) & ((
		patterns: string | readonly string[],
		options?: globby.GlobbyOptions
	) => string[])
```

alex-page:
Getting a similar error

```
(alias) globbySync(patterns: string | readonly string[], options?: {
    stats: true;
} & globby.GlobbyOptions): Stats[] (+1 overload)
import globbySync
Argument of type 'Stats[]' is not assignable to parameter of type 'string[]'.ts(2345)
```

olup:
I am sorry where and how do you get this error ? I just tested it myself and no tests are down. I might propose a PR myself if you are ok with this ?

olup:
You'd better actually omit `stats` key from the options : 
```
	sync: ((
		patterns: string | readonly string[],
		options: Except<globby.GlobbyOptions, 'stats'> & { stats: true }
	) => Stats[]) & ((
		patterns: string | readonly string[],
		options?: globby.GlobbyOptions
	) => string[]);
```

olup:
Sorry, to be absolutely fair, this is not the `Except` bit that does the trick but **removing the optional for the option key**. This is normal: otherwise, your overload could have options as undefined and then is not distinguishable from the other type. So final code is this : 
```
	sync: ((
		patterns: string | readonly string[],
		options: globby.GlobbyOptions & { stats: true }
	) => Stats[]) & ((
		patterns: string | readonly string[],
		options?: globby.GlobbyOptions
	) => string[]);
```

alex-page:
This is awesome. I have it working for `sync` but the `Promise<>` is a little bit different.

olup:
You can check my pr, It resolves both cases.

PR #174, opened 2021-04-13
Add types for stats option's overload

olup:
This PR adds the stats return type when the stats option is enabled. It overseeds https://github.com/sindresorhus/globby/pull/173.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,10 +1,11 @@
 import {Options as FastGlobOptions} from 'fast-glob';
+import {Stats} from 'fs';
 
 declare namespace globby {
 	type ExpandDirectoriesOption =
 		| boolean
 		| readonly string[]
-		| {files?: readonly string[]; extensions?: readonly string[]};
+		| { files?: readonly string[]; extensions?: readonly string[] };
 
 	interface GlobbyOptions extends FastGlobOptions {
 		/**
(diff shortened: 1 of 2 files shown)


alex-page (approved):
Thanks @olup

olup:
One question tho, is the stat option really returning objects of type `Stats` ?

alex-page:
Yes because it has all of the different file stats if you select that value.

index.d.ts
```diff
@@ -1,10 +1,11 @@
 import {Options as FastGlobOptions} from 'fast-glob';
+import {Stats} from 'fs';
 
 declare namespace globby {
 	type ExpandDirectoriesOption =
 		| boolean
 		| readonly string[]
-		| {files?: readonly string[]; extensions?: readonly string[]};
+		| { files?: readonly string[]; extensions?: readonly string[] };
```
sindresorhus, line 8: Don't do unrelated changes.

index.d.ts
```diff
@@ -86,12 +87,15 @@ declare const globby: {
...
 	@param patterns - See the supported [glob patterns](https://github.com/sindresorhus/globby#globbing-patterns).
 	@param options - See the [`fast-glob` options](https://github.com/mrmlnc/fast-glob#options-3) in addition to the ones in this package.
-	@returns The matching paths.
+	@returns The matching paths - or the matching path's stats if the `stats` option is set.
 	*/
-	sync: (
+	sync: ((
+		patterns: string | readonly string[],
+		options: globby.GlobbyOptions & { stats: true }
+	) => Stats[]) & ((
```
sindresorhus, line 95: Why is this not just an overload?

sindresorhus (changes requested):

index.d.ts
olup: Sorry, this is probably auto format. You don't seem to provide a prettier config or eslintrc ?

index.d.ts
olup: I am not sure but as this is declare as a key of a const object, you cannot just re-declare the key. There is a basic overload declared as the `globby` global result, but for the sync method, this is the best I could come up with.

index.d.ts
sindresorhus: > You don't seem to provide a prettier config or eslintrc ?

No, but you should default to not formatting if such configs do not exist.

index.d.ts
sindresorhus: Alright. You still need to fix the formatting.

olup: Hum alright, I'll do that on Monday...

Purpzie: Perhaps it could be changed into a regular overload? I'm unsure if it would be treated differently that way.

Apologies for butting in, I'm also editing this file for #177 and ran into this problem as well.

olup: Sorry for the delay, better ?

PR #178, opened 2021-06-15
Support `objectMode` option in the TypeScript types

### LINKED ISSUE #177: Types: `string[]` is still assumed to be returned even if the `objectMode` option is true
I would submit a pull request, but I'm new to Typescript and I'm not sure how to define this. Currently being forced to do this:

```ts
type GlobbyResult = { name: string, path: string, dirent: Dirent }[];
const results: GlobbyResult = await globby([/* globs */], { objectMode: true }]) as any;
```

Thank you in advance!

Purpzie:
Fixes #177
Closes #174

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,10 +1,12 @@
-import {Options as FastGlobOptions} from 'fast-glob';
+import {Options as FastGlobOptions, Entry as FastGlobEntry} from 'fast-glob';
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Type tests are failing

Purpzie:
Oh woops. Forgot that more specific overloads should be specified before more generic ones.

PR #180, opened 2021-07-07
adjusting catch syntax

danielsasaki87:
In ES5 it won't compile


sindresorhus:
This package doesn't support ES5.

PR #188, opened 2021-08-17
Allow node ^15

dmetzner:
v12.0.1 had breaking changes @sindresorhus


sindresorhus:
Node.js 15 was already implicitly unsupported as it's not supported by the Node.js team anymore.

---

> After six months, odd-numbered releases (9, 11, etc.) become unsupported - https://nodejs.org/en/about/releases/

PR #194, opened 2021-09-24
Give notice about Window support

bradenhs:
I wish I had time to dig into the problem here and create a PR! In lieu of that more ideal contribution I thought it prudent to at least give folks notice that this library has less than ideal Windows support. When making decisions about libraries to use this notice will prove helpful to folks and hopefully prevent downstream issues (here's an example of one place this has popped up https://github.com/grafana/grafana/issues/35674). It's my sincere hope that this notice will get reverted in the near future! Really appreciate all your work Sindre!


sindresorhus:
This is not useful without any context. There are already open issue about improving Windows support.

PR #196, opened 2021-11-03
Fix node imports

Tenpi:
Fix the node imports which are broken on newer versions of node

E.g. `Error: Cannot find module node:util`

Node version: 16.3.0


sindresorhus:
These are completely valid imports: https://github.com/nodejs/node/issues/38343

Tenpi:
The imports are not working for me, if it's not a Node issue then it might be Typescript or Electron.

For compatibility sake I suggest that you use the traditional imports, otherwise I will have to stick to using my fork of this project.

PR #201, opened 2022-01-12
Support `URL` as `cwd`

### LINKED ISSUE #176: Accept `URL` as `options.cwd`?
```js
import globby from 'globby';
const dir = new URL('./src', import.meta.url);

console.log(await globby('*', {cwd: dir}));

// TypeError: Expected `cwd` to be of type `string` but received type `object`
```

Currently, I have to

```diff
+ import {fileURLToPath} from 'node:url';
  import globby from 'globby';
  const dir = new URL('./src', import.meta.url);

- console.log(await globby('*', {cwd: dir}));
+ console.log(await globby('*', {cwd: fileURLToPath(dir)}));
```

fisker:
I build [`url-or-path`](https://github.com/fisker/url-or-path) few months ago, it actually accept `path`, `URL` and `URL.href`, should we add test for `URL.href` too?

Fixes #176

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -12,7 +12,9 @@ export type ExpandDirectoriesOption =
 	| readonly string[]
 	| {files?: readonly string[]; extensions?: readonly string[]};
 
-export interface Options extends FastGlobOptions {
+type FastGlobOptionsWithoutCwd = Omit<FastGlobOptions, 'cwd'>;
+
+export interface Options extends FastGlobOptionsWithoutCwd {
 	/**
 	If set to `true`, `globby` will automatically glob directories for you. If you define an `Array` it will only glob files that matches the patterns inside the `Array`. You can also define an `Object` with `files` and `extensions` like in the example below.
 
@@ -43,10 +45,17 @@ export interface Options extends FastGlobOptions {
 	@default false
 	*/
 	readonly gitignore?: boolean;
+
+	/**
+	The current working directory in which to search.
+
+	@default process.cwd()
+	*/
+	readonly cwd?: URL | string;
 }
 
 export interface GitignoreOptions {
-	readonly cwd?: string;
+	readonly cwd?: URL | string;
 	readonly ignore?: readonly string[];
 }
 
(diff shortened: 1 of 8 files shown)


sindresorhus:
I think Node.js should expose a method for this. It's such a common need now that everything is starting to support `URL`. Maybe you wanna open an issue about it on the Node.js repo?

fisker:
Removed dependency, and added test for `file:` string too.

fisker:
> It's such a common need now that everything is starting to support URL. Maybe you wanna open an issue about it on the Node.js repo?

I think this is a temporary need. Once everything supports `URL`, convert to path won't be necessary.

sindresorhus:
> I think this is a temporary need. Once everything supports URL, convert to path won't be necessary. Anyway,

No, there will always be a need as your argument only works if you directly pass the input to a Node.js API. Many packages manipulate the paths somehow and dealing with path and not URL is much nicer then.

index.d.ts
```diff
@@ -43,10 +45,17 @@ export interface Options extends FastGlobOptions {
...
+	The current working directory in which to search.
+
+	@default process.cwd()
+	*/
+	readonly cwd?: URL | string;
 }
 
 export interface GitignoreOptions {
-	readonly cwd?: string;
+	readonly cwd?: URL | string;
```
antongolub, line 58: `URL` interface belongs to `lib.dom.d.ts`, so now we force TS users to inject `compilerOptions.lib: ['dom']` to resolve it. It breaks TS builds and formally this is a breaking change.

sindresorhus: https://nodejs.org/api/url.html#the-whatwg-url-api

antongolub: So it should be imported in `index.d.ts`. It does not belong to global by default.

antongolub: @sindresorhus 
I really don't want to fix hundreds of our `globby@^12`-dependant projects all night )). Could you plz roll the fix #206?

fisker: I'm sorry for breaking your projects, but I know nothing about the typing, I can't help.

PR #205, opened 2022-01-14
Fix code example on readme

EarthlingDavey:
It seems `import {globby} from 'globby';` should be `import globby from 'globby';` so I have removed the `{}`s.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -27,7 +27,7 @@ $ npm install globby
 ```
 
 ```js
-import {globby} from 'globby';
+import globby from 'globby';
 
 const paths = await globby(['*', '!cake']);
 


fisker:
The original one is correct, it's named export.

EarthlingDavey:
ofc there must be some reason that only `import globby from 'globby';` works with my setup.
Seems like you don't really mind. And I don't mind either :)
See ya 👎

fisker:
I'm not the repo owner, the export is here https://github.com/sindresorhus/globby/blob/de4082b09dcd36f7053a4e63d3dff8994dee35e8/index.js#L123

EarthlingDavey:
ok sure thing. IDK why it would only play nice in my nextjs project without the `{}`

PR #206, opened 2022-01-16
Work around TypeScript type problem with `URL` global

antongolub:
relates #201

https://github.com/sindresorhus/globby/pull/201#discussion_r785471096

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,4 @@
-import {URL} from 'node:url';
+import {URL} from 'node:url'; // TODO: Remove this when https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34960 is fixed.
 import {Options as FastGlobOptions, Entry} from 'fast-glob';
 
 export type GlobEntry = Entry;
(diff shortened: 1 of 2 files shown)


sindresorhus:
This should not be needed. `URL` is a valid global in Node.js.

antongolub:
Correct, should not be needed. However built-in TS libdefs (4.5.4) does not reflect this fact.

image

 And `@types/node` too.
image

antongolub:
I'll make a PR for TS/definetlyTyped, but right now I'd like to fix our CI/CD.

antongolub:
Clarified for myself what's going on: it seems if we have no explicit `lib` declaration in `compilerOptions`, TS attaches `dom`

image

sindresorhus:
I forgot, this is actually a long running TS issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34960 (I cannot believe they have still not fixed that...)

sindresorhus:
https://github.com/sindresorhus/globby/releases/tag/v12.2.0

PR #207, opened 2022-01-17
Accept `URL` in function returned by `isGitIgnored`

### LINKED ISSUE #202: Accept `URL` in function returns by `isGitIgnored()` and `isGitIgnoredSync()`
This is a little different from #176, I didn't implement in  #201

```diff
const isIgnored = await isGitIgnored();

- console.log(isIgnored(fileURLToPath(url)));
+ console.log(isIgnored(url));
```

fisker:
Fixes #202

I just realized this function also accept object with `path`, I guess it's easier to use in `index.js`, but as a public method, I don't think it should do such thing, anyway I didn't touch it.

### DIFF
--- a/gitignore.js
+++ b/gitignore.js
@@ -56,7 +56,7 @@ const ensureAbsolutePathForCwd = (cwd, p) => {
 	return path.join(cwd, p);
 };
 
-const getIsIgnoredPredicate = (ignores, cwd) => p => ignores.ignores(slash(path.relative(cwd, ensureAbsolutePathForCwd(cwd, p.path || p))));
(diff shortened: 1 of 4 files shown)


sindresorhus:
You need to update the readme too.

fisker:
Ha, you already did..  but  you missed one place.

fisker:
The object with `path` property is also not mentioned there, we should remove support for it.

sindresorhus:
> Ha, you already did.. but you missed one place. isGitIgnored section

Fixed in `main`.

> The object with path property is also not mentioned there, we should remove support for it.

👍

PR #208, opened 2022-01-17
Drop support for checking object with `path` property in function returns by `isGitIgnored`

fisker:
Ref: https://github.com/sindresorhus/globby/pull/207#issuecomment-1014263028

This behavior is never documented.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,7 +34,7 @@ const checkCwdOption = options => {
 	}
 };
 
-const getPathString = p => p.stats instanceof fs.Stats ? p.path : p;
+const getPathString = fastGlobResult => fastGlobResult.path || fastGlobResult;
 
 export const generateGlobTasks = (patterns, taskOptions) => {
 	patterns = arrayUnion([patterns].flat());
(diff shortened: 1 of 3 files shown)


fisker:
I found other bugs...

index.js
```diff
@@ -34,7 +34,7 @@ const checkCwdOption = options => {
 	}
 };
 
-const getPathString = p => p.stats instanceof fs.Stats ? p.path : p;
+const getPathString = fastGlobResult => fastGlobResult.path || fastGlobResult;
```
fisker, line 37: This function mean to handle fastGlobResult, but `p.stats` is not always true, see  [`objectMode` option in `fast-glob`](https://github.com/mrmlnc/fast-glob#objectmode)

index.js
```diff
@@ -152,7 +152,7 @@ export const globby = async (patterns, options = {}) => {
 	const [filter, tasks] = await Promise.all([getFilter(options), getTasks()]);
 	const paths = await Promise.all(tasks.map(task => fastGlob(task.pattern, task.options)));
 
-	return arrayUnion(...paths).filter(path_ => !filter(getPathString(path_)));
```
fisker, line 145: `filter` already handles `objectMode` result, not needed here, `globbySync` and `globbyStream` also not doing this.

fisker:
This one is ready, but I saw another bug here, the `arrayUnion` use is not correct, when use `fastGlob` with `objectMode`, `arrayUnion` can't work, we should implement a function to unique the fastGlob result with `path`, I'll send another PR.

sindresorhus:
Thanks for catching these bugs. I really appreciate it.

PR #211, opened 2022-01-17
Remove `array-union`

fisker:
After we merge #210, we should be able to remove this package.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,4 @@
 import fs from 'node:fs';
-import arrayUnion from 'array-union';
 import merge2 from 'merge2';
 import fastGlob from 'fast-glob';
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -146,46 +163,38 @@ export const globby = async (patterns, options = {}) => {
 			return Promise.all(globs.map(globToTask(task)));
 		}));
 
-		return arrayUnion(...tasks);
```
fisker, line 149: Tasks are all objects, `arrayUnion` not doing anything here.

fisker:
Random failure? I can't reprod on Windows + Node.js 14.18.3

fisker:
Can you rerun?

PR #215, opened 2022-01-18
Simplify task expansion

fisker:
- Merge expand `pattern` and expand `ignore` into one step
- Skip expand when `options.expandDirectories` is `false`, we calling  `Array#map()` unnecessarily before


fisker:
There's another thing to do after this. We are normalizing `expandDirectories` repeatedly.

PR #216, opened 2022-01-18
Improve test coverage

fisker:
When I work on #215, I found that `expandDirectories` not even tested in promise version, I added a function to run all 3 methods in every tests, make sure they have the same result, this also simplify the tests.


sindresorhus:
Great!

PR #217, opened 2022-01-18
Minor refactoring

fisker:
- Move `toPath` and `FilterStream` to utils.js
- Remove `ObjectTransform` class, there was another class extends it, but not any more
- Use really private property


utils.js
```diff
@@ -0,0 +1,31 @@
...
+export const toPath = urlOrPath => {
+	if (!urlOrPath) {
+		return urlOrPath;
+	}
+
+	if (urlOrPath instanceof URL) {
+		urlOrPath = urlOrPath.href;
+	}
+
+	return urlOrPath.startsWith('file://') ? fileURLToPath(urlOrPath) : urlOrPath;
```
sindresorhus, line 13: I just realized, we shouldn't do `urlOrPath.startsWith('file://')`. Only the instanceof check. We should only accept `URL` or string path.

package.json
```diff
@@ -23,8 +23,7 @@
 		"index.js",
 		"index.d.ts",
 		"gitignore.js",
-		"stream-utils.js",
-		"to-path.js"
+		"utils.js"
```
sindresorhus, line 26:
```suggestion
		"utilities.js"
```

utilities.js
```diff
@@ -0,0 +1,27 @@
+import {fileURLToPath} from 'node:url';
+import {Transform} from 'node:stream';
+
+export const toPath = urlOrPath => {
+	if (!urlOrPath) {
+		return urlOrPath;
+	}
+
+	return urlOrPath instanceof URL ? fileURLToPath(urlOrPath) : urlOrPath;
+};
```
sindresorhus, line 10:
```suggestion
export const toPath = urlOrPath => urlOrPath instanceof URL ? fileURLToPath(urlOrPath) : urlOrPath;
```

utilities.js
```diff
@@ -0,0 +1,27 @@
...
+		super({objectMode: true});
+		this.#filter = filter;
+	}
+
+	_transform(data, encoding, callback) {
+		if (this.#filter(data)) {
+			this.push(data);
+		}
+
+		callback();
```
sindresorhus, line 25:
```suggestion
		callback(undefined, this.#filter(data) && data);
```

fisker:
👍

fisker:
https://github.com/nodejs/node/blob/37f1dd9ccdee8c54e1a3465f73e727eb87555fe5/lib/internal/streams/transform.js#L191

fisker:
`this.#filter(data) && data` = `false`

fisker:
Just saw there is a `transform` option.

PR #219, opened 2022-01-19
Improve task expansion

fisker:
`fast-glob` accepts array as patterns, so we don't need flat tasks.

I'm not sure if this improves performance, but I don't think it will be slower.

In `generateGlobTasks` we can also avoid use multiple task, but I don't if worth to improve that part.

The benchmark test is broken, maybe I can tell if we fix that?


sindresorhus:
And if it's slower, it should ideally be fixed in `fast-glob`, so it benefits everyone.

sindresorhus:
> The benchmark test is broken, maybe I can tell if we fix that?

Feel free to fix it, but simple benchmarks don't usually tell the full story. It only focuses on one specific performance case.

PR #220, opened 2022-01-19
Fix `bench` script

fisker:


fisker:
#219 did improve performance a little bit.

PR #223, opened 2022-01-20
Don't pass `options.ignore` to `isGitIgnored`

fisker:
`ignore` option in `isGitIgnored()` mean to ignore `.gitignore` files,  but when running `globby()`, `ignore` mean to ignore matched files.

For example:

```js
globbySync('**/*', {gitignore: true, ignore: ['.gitignore']})
```

This mean to glob all files that's not `gitignore`d, but except `.gitignore` file, not mean don't read the `.gitignore` file.

Before:

```text
> globbySync('**/*', {gitignore: true, ignore: ['.gitignore']}).filter(file => file.startsWith('node_modules')).length
11662
```

After:

```text
> globbySync('**/*', {gitignore: true, ignore: ['.gitignore']}).filter(file => file.startsWith('node_modules')).length
0
```

This is a little hard to test, I didn't add.

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,10 +53,10 @@ const normalizeArguments = fn => async (patterns, options) => fn(toPatternsArray
 const normalizeArgumentsSync = fn => (patterns, options) => fn(toPatternsArray(patterns), normalizeOptions(options));
 
 const getFilter = async options => createFilterFunction(
-	options.gitignore && await isGitIgnored({cwd: options.cwd, ignore: options.ignore}),
+	options.gitignore && await isGitIgnored({cwd: options.cwd}),
 );
 const getFilterSync = options => createFilterFunction(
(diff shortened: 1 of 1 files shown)


fisker:
@sindresorhus After second thought, this `ignore` option for `isGitIgnored` doesn't make sense to me either, why we don't want to skip reading a specific `.gitignore` file and let the result be `true`?

Do you think we should remove it?

sindresorhus:
No idea. It was added a long time ago. I do agree with your reasoning though.

fisker:
It was added in https://github.com/sindresorhus/globby/pull/48, I think it's added for `xo`, maybe mean to check file is ignored by `ignore` option, but it never works that way.

sindresorhus:
Yeah, let's remove it.

PR #224, opened 2022-01-20
Improve `isGitIgnored`

fisker:


gitignore.js
```diff
@@ -7,9 +7,9 @@ import slash from 'slash';
 import {toPath} from './utilities.js';
 
 const DEFAULT_IGNORE = [
-	'**/node_modules/**',
-	'**/flow-typed/**',
-	'**/coverage/**',
+	'**/node_modules',
+	'**/flow-typed',
+	'**/coverage',
 	'**/.git',
```
fisker, line 13: @mrmlnc 

Is `dir` and `dir/**` in `ignore` option behave different?

[This explanation](https://github.com//fast-glob#how-to-exclude-directory-from-reading) says they are different when writing `negative pattern`, but not clear to me if it makes difference when the patterns used in `ignore` option, I assume we should prefer `dir` over `dir/**` in `ignore` option too?

fisker:
Close in favor of #225

PR #225, opened 2022-01-21
Remove `ignore` option for `isGitIgnored` and `isGitIgnoredSync`

fisker:
Ref https://github.com/sindresorhus/globby/pull/223#issuecomment-1017255797

### DIFF
--- a/gitignore.js
+++ b/gitignore.js
@@ -6,13 +6,6 @@ import gitIgnore from 'ignore';
 import slash from 'slash';
 import {toPath} from './utilities.js';
 
-const DEFAULT_IGNORE = [
-	'**/node_modules/**',
-	'**/flow-typed/**',
-	'**/coverage/**',
-	'**/.git',
-];
-
 const mapGitIgnorePatternTo = base => ignore => {
 	if (ignore.startsWith('!')) {
 		return '!' + path.posix.join(base, ignore.slice(1));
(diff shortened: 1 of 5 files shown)


gitignore.js
```diff
@@ -6,13 +6,6 @@ import gitIgnore from 'ignore';
 import slash from 'slash';
 import {toPath} from './utilities.js';
 
-const DEFAULT_IGNORE = [
-	'**/node_modules/**',
-	'**/flow-typed/**',
-	'**/coverage/**',
-	'**/.git',
-];
```
sindresorhus, line 14: While we're removing the ignore option for `isGitIgnored`, I think `globby` should still ignore these paths by default in the main globby methods.

fisker: These were ONLY used  when glob `.gitignore` file, changing `globby()` to ignore files in these directories is a new breaking change to the `globby()` function, it out scope of this PR.

And if we decide to ignore them by default, but how to unignore them?

fisker: Ah, you are right, misunderstood.

gitignore.js
```diff
@@ -6,12 +6,15 @@ import gitIgnore from 'ignore';
...
+const gitignoreGlobOptions = {
+	ignore: [
+		'**/node_modules',
+		'**/flow-typed',
+		'**/coverage',
+		'**/.git',
```
fisker, line 14: As I tested, when use `**/node_modules/` and `**/node_modules/**/*`, `fast-glob` will read those directories, but when use `**/node_modules` , `fast-glob` won't read them.

fisker: Answering this question https://github.com/sindresorhus/globby/pull/224#discussion_r788528852

fisker:
I still have one thing want to do before you can release, the `expandDirectories` requires file access, though I already reduced the tasks, still expanding `ignore` repeatedly, we can expand them in 0~2 steps, one for the common `ignore`, one for the negative patterns.

fisker:
I'm going to give up for this https://github.com/sindresorhus/globby/pull/225#issuecomment-1018440577.

I have tried, the logic became more complex, but didn't see much performance improvement, maybe because the file access is fast enough. Adding this logic also only benefit cases with multiple discontinuous negative patterns, it's not very common.

sindresorhus:
Thanks for all the work you have been doing. I did a new release: https://github.com/sindresorhus/globby/releases/tag/v13.0.0

PR #231, opened 2022-02-01
Fix duplicated result when using globstar

### LINKED ISSUE #175: Duplicated results when directory specified with globstar
Here's a situation that I've come across. There is a directory `a` that has a single file in it, `a/b`. The following program shows that `globby` will return the file twice (once with a leading `./` and once without. However, `fast-glob` only returns this file once. 

```js
const g = require('globby')
const f = require('fast-glob')

console.log(g.sync(['./a','./a/**'])) // [ 'a/b', './a/b' ]
console.log(f.sync(['./a','./a/**'])) // [ './a/b' ]
```

When this is run without the leading `./` both `globby` and `fast-glob` return the file only once. I think `fast-glob` is the correct behaviour.

Using v11.0.3 of globby. And v3.2.5 of fast-glob.

jopemachine:
Fixes #175.

I updated `createFilterFunction` using normalized path string instead of just path.

So, the below code lines return the commented values instead of `[ 'a/b', './a/b' ]`.

```js
// below code returns [ './a/b' ]
console.log(globbySync(['./a/**', './a']));

// below code returns [ 'a/b' ]
console.log(globbySync(['./a', './a/**']));
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,4 +1,5 @@
 import fs from 'node:fs';
+import nodePath from 'node:path';
 import merge2 from 'merge2';
 import fastGlob from 'fast-glob';
 import dirGlob from 'dir-glob';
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can you add a test?

PR #232, opened 2022-02-16
Refactor with `gensync`

fisker:
I found [`gensync`](https://github.com/loganfsmyth/gensync) is a very handy module, it allows you to write the async and sync logic in one function.

How do you think?

### DIFF
--- a/ignore.js
+++ b/ignore.js
@@ -1,7 +1,7 @@
 import process from 'node:process';
 import fs from 'node:fs';
 import path from 'node:path';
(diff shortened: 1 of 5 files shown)


sindresorhus:
Not a big fan, to be honest. I find the code less readable after this change, and the existing duplication is not that bad. This also introduces a dependency, and a risk, as `gensync` is a big hack. All this to save 13 lines of code. Do you think it's worth it?

fisker:
Okay.

PR #242, opened 2022-12-09
Fixes exception in a top path edge case

voxpelli:
The gitignore is throwing an exception when dealing with absolute paths if that path is the very same as the working directory.

This is the case when both `expandDirectories: false` and `onlyFiles: false` are set and the input is `.` and the `cwd` is set to an absolute path.

I ran across this when using `globby` in a project of mine and I reproduced it with a test case here and then fixed it.

### DIFF
--- a/ignore.js
+++ b/ignore.js
@@ -52,7 +52,7 @@ const getIsIgnoredPredicate = (files, cwd) => {
 	return fileOrDirectory => {
 		fileOrDirectory = toPath(fileOrDirectory);
 		fileOrDirectory = toRelativePath(fileOrDirectory, cwd);
-		return fileOrDirectory && ignores.ignores(slash(fileOrDirectory));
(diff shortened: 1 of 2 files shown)


ignore.js
```diff
@@ -52,7 +52,7 @@ const getIsIgnoredPredicate = (files, cwd) => {
 	return fileOrDirectory => {
 		fileOrDirectory = toPath(fileOrDirectory);
 		fileOrDirectory = toRelativePath(fileOrDirectory, cwd);
-		return ignores.ignores(slash(fileOrDirectory));
+		return fileOrDirectory && ignores.ignores(slash(fileOrDirectory));
```
fisker, line 55: Let's wrap the result with `Boolean` so it can match the type definition?

voxpelli: @fisker Fixed + I also fixed the tests that were failing on windows

fisker (approved):

PR #246, opened 2023-04-07
Fix error when reading inaccessible directories with `gitignore: true` and `suppressErrors: true`

fwouts:
Currently, using `gitignore: true` with `suppressErrors: true` still shows an error if globby encounters, for example, a directory without read access.

See the following repo for an example: https://github.com/fwouts/globby-eaccess-bug

This fixes the issue and adds a corresponding test. Other tests that rely on the same temporary directory are made serial so there is no race condition between them, as this new test changes the layout of that directory.

Let me know if any further adjustments are required :)


sindresorhus:
Instead of making many of the tests serial, I would prefer to use a different temp directory. You could use https://github.com/sindresorhus/tempy

tests/globby.js
```diff
@@ -261,6 +261,23 @@ test('gitignore option and objectMode option', async t => {
...
+test.serial('gitignore option and suppressErrors option', async t => {
+	const fooPath = path.join(temporary, 'foo.tmp');
+	const notignoredPath = path.join(temporary, 'notignored');
+	try {
+		fs.mkdirSync(fooPath);
+		fs.chmodSync(fooPath, 0o000);
+		fs.writeFileSync(notignoredPath, 'notignored', 'utf8');
+		const result = await runGlobby(t, temporary, {gitignore: true, suppressErrors: true});
+		t.is(result.length, 1);
+		t.truthy(result.includes('tmp/notignored'));
```
sindresorhus, line 273: The test will pass even if something throws in this block. Use `t.plan()` to ensure the assertions are executed.

fwouts: I didn't quite get this, as there's no `catch` block, but I removed the cleanup since we're now using a real temporary directory instead of the local `tmp/` subdirectory. Let me know if that works for you.

fwouts:
> Instead of making many of the tests serial, I would prefer to use a different temp directory. You could use https://github.com/sindresorhus/tempy

Sounds good, done.

sindresorhus:
Looks like tests are failing on Windows.

PR #259, opened 2023-11-12
Fix read permission error on ignore files search

### LINKED ISSUE #254: `.gitignore: true` makes globby crash on directory with bad read permissions
In our project we have a folder under the root folder which is .gitignored and used by a docker image to write to database data.
Since it's running inside a Docker container the directory has a different owner and the files can not be read by the user outside of the docker container.

When running globby with the `gitignore: true` option it crashes.

### Steps to reproduce

```sh
mkdir someproject
cd someproject
npm init -y && npm i globby
mkdir testdir
chmod -r testdir # remove read access
node --input-type=module -e "const { globby } = await import('globby'); console.log(await globby('test', { gitignore: true }))"
```

The last line crashes with an error like this:
```
[Error: EACCES: permission denied, scandir '/Users/d441365/projects/deedop/freigeist/someproject/testdir'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path: '/Users/d441365/projects/deedop/freigeist/someproject/testdir'
}
```

Setting `gitignore: false` makes the error go away

It's not globby's fault that it can't read the directory. Would you consider adding a way to alter globby's behavior in this case though? For example with an option like `unreadable: 'skip'` which would simply skip directories that can't be read.

PS the whole thing came up when using `knip` which is using globby under the hood https://github.com/webpro/knip/issues/172

### LINKED ISSUE #258: Prevent search of ignore files in ignored files list
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `globby@13.2.2` for the project I'm working on.

The issue that gave birth to this fix can be found in [`xo`’s repository](https://github.com/xojs/xo/issues/737). You’ll find there all the explanations about why I did need it.

While I believe this is not a great fix (code-wise), it does solve my issue. If you want me to open a PR, I’ll gladly do so! Please let me know :slightly_smiling_face: 

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/globby/ignore.js b/node_modules/globby/ignore.js
index 3d8e1a8..74de387 100644
--- a/node_modules/globby/ignore.js
+++ b/node_modules/globby/ignore.js
@@ -60,12 +60,13 @@ const normalizeOptions = (options = {}) => ({
 	cwd: toPath(options.cwd) || process.cwd(),
 	suppressErrors: Boolean(options.suppressErrors),
 	deep: typeof options.deep === 'number' ? options.deep : Number.POSITIVE_INFINITY,
+  ignore: Array.isArray(options.ignore) ? [...options.ignore, ...ignoreFilesGlobOptions.ignore] : ignoreFilesGlobOptions.ignore
 });
 
 export const isIgnoredByIgnoreFiles = async (patterns, options) => {
-	const {cwd, suppressErrors, deep} = normalizeOptions(options);
+	const {cwd, suppressErrors, deep, ignore} = normalizeOptions(options);
 
-	const paths = await fastGlob(patterns, {cwd, suppressErrors, deep, ...ignoreFilesGlobOptions});
+	const paths = await fastGlob(patterns, {cwd, suppressErrors, deep, ...ignoreFilesGlobOptions, ignore});
 
 	const files = await Promise.all(
 		paths.map(async filePath => ({
@@ -78,9 +79,9 @@ export const isIgnoredByIgnoreFiles = async (patterns, options) => {
 };
 
 export const isIgnoredByIgnoreFilesSync = (patterns, options) => {
-	const {cwd, suppressErrors, deep} = normalizeOptions(options);
+	const {cwd, suppressErrors, deep, ignore} = normalizeOptions(options);
 
-	const paths = fastGlob.sync(patterns, {cwd, suppressErrors, deep, ...ignoreFilesGlobOptions});
+	const paths = fastGlob.sync(patterns, {cwd, suppressErrors, deep, ...ignoreFilesGlobOptions, ignore});
 
 	const files = paths.map(filePath => ({
 		filePath,
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

nicolas-goudry:
This PR should fix #254 and fix #258. As well as `xo`’s [#737](https://github.com/xojs/xo/issues/737) when `globby`’s dependency gets upgraded.

It allows to forward the `ignore` option field passed to `globby`’s `isIgnoredByIgnoreFiles` to `fast-glob`.

To give some more context, this function is used to find all `.gitignore` files relative to the current working directory. If any subdirectory of the current working directory is not readable, `fast-glob` exits with an error.

By forwarding the `ignore` option to `fast-glob` when searching for ignore files, we prevent this behavior.

A few things that should be considered before this gets merged:

**Documentation**

There are currently no docs about this, but since the readme file mentions that the `options` object extends the `fast-glob` options, I’m not sure if it’s worth to document.

**Tests**

When a directory is not readable, the behavior of `fast-glob` is to **exit**, not to _throw_.
So, if we remove the `ignore` option of the test, `ava` will fail **all** tests run with the following error:

```
Exiting due to process.exit() when running tests/ignore.js
```

Some side nodes about the test:
- `.notThrowsAsync` is used, but as said above: `fast-glob` doesn’t throw, it exits ; so this might be wrong. You tell me
- `.serial` is used to prevent other tests to fail because of the bad directory permissions
- `.teardown` is used to revert the directory permissions to something readable

### DIFF
--- a/ignore.js
+++ b/ignore.js
@@ -8,13 +8,13 @@ import slash from 'slash';
 import {toPath} from 'unicorn-magic';
 import {isNegativePattern} from './utilities.js';
 
+const defaultIgnoredDirectories = [
+	'**/node_modules',
+	'**/flow-typed',
+	'**/coverage',
+	'**/.git',
+];
 const ignoreFilesGlobOptions = {
-	ignore: [
-		'**/node_modules',
-		'**/flow-typed',
-		'**/coverage',
-		'**/.git',
-	],
 	absolute: true,
 	dot: true,
 };
@@ -62,12 +62,13 @@ const normalizeOptions = (options = {}) => ({
 	cwd: toPath(options.cwd) ?? process.cwd(),
 	suppressErrors: Boolean(options.suppressErrors),
 	deep: typeof options.deep === 'number' ? options.deep : Number.POSITIVE_INFINITY,
+	ignore: [...options.ignore ?? [], ...defaultIgnoredDirectories],
 });
 
 export const isIgnoredByIgnoreFiles = async (patterns, options) => {
-	const {cwd, suppressErrors, deep} = normalizeOptions(options);
+	const {cwd, suppressErrors, deep, ignore} = normalizeOptions(options);
 
-	const paths = await fastGlob(patterns, {cwd, suppressErrors, deep, ...ignoreFilesGlobOptions});
+	const paths = await fastGlob(patterns, {cwd, suppressErrors, deep, ignore, ...ignoreFilesGlobOptions});
 
 	const files = await Promise.all(
 		paths.map(async filePath => ({
@@ -80,9 +81,9 @@ export const isIgnoredByIgnoreFiles = async (patterns, options) => {
 };
 
 export const isIgnoredByIgnoreFilesSync = (patterns, options) => {
-	const {cwd, suppressErrors, deep} = normalizeOptions(options);
+	const {cwd, suppressErrors, deep, ignore} = normalizeOptions(options);
 
-	const paths = fastGlob.sync(patterns, {cwd, suppressErrors, deep, ...ignoreFilesGlobOptions});
+	const paths = fastGlob.sync(patterns, {cwd, suppressErrors, deep, ignore, ...ignoreFilesGlobOptions});
 
 	const files = paths.map(filePath => ({
 		filePath,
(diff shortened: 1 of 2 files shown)


nicolas-goudry:
To fix #254 with the provided steps to reproduce, would give the following:

```
mkdir someproject
cd someproject
npm init -y && npm i globby
mkdir testdir
chmod -r testdir # remove read access
node --input-type=module -e "const { globby } = await import('globby'); console.log(await globby('test', { gitignore: true, ignore: ['testdir'] }))"
```

nicolas-goudry:
@sindresorhus could you please take a look at this?

PR #267, opened 2024-05-14
Do not ignore pattern start with `./`

zanminkian:
This PR fix 2 issues:

### Issue 2: pattern starts with `./` or `./` should not be ignored

```
foo
├── .gitignore ---> content: ./bar.js
└── bar.js ---> should not be ignored, but actually ignored by globby
```

### DIFF
--- a/fixtures/gitignore/.gitignore
+++ b/fixtures/gitignore/.gitignore
@@ -1,2 +1,3 @@
 foo.js
 !bar.js
+./baz.js
--- a/ignore.js
+++ b/ignore.js
@@ -30,7 +30,7 @@ const parseIgnoreFile = (file, cwd) => {
 
 	return file.content
 		.split(/\r?\n/)
-		.filter(line => line && !line.startsWith('#'))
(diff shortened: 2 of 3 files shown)


fisker:
>   - Handle trailing whitespaces:
>    - `'a '`(one space) should not match `'a  '`(two spaces).
>    - `'a \ '` matches `'a  '

Quote from https://github.com/kaelzhang/node-ignore/tree/master?tab=readme-ov-file#why-another-ignore. It seems spaces are useful.

zanminkian:
I'm not pretty sure about issue 1 so I updated this PR.

sindresorhus:
Thanks for the PR! However, after thorough testing, I found that the current implementation already handles patterns starting with `./` or `../` correctly, matching Git's behavior.

The `ignore` library (which globby uses internally) treats patterns starting with `./` as literal patterns that don't match normalized paths, so no code changes are needed.

I've added a test case to document this behavior and prevent future confusion.

PR #269, opened 2024-09-10
Export `isIgnoredByIgnoreFiles` and `isIgnoredByIgnoreFilesSync` functions

jiawei397:
I am developing a feature to dynamically monitor folder changes. There is a file similar to `. gitignore` that needs to determine whether it is ignored after the file adds. This requires the use of the underlying methods `isIgnoredByIgnoreFiles` or `isIgnoredByIgnoreFilesSync` that are not exposed.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -207,39 +207,39 @@ export function isGitIgnoredSync(options?: GitignoreOptions): GlobbyFilterFuncti
 export function convertPathToPattern(source: string): FastGlob.Pattern;
 
 /**
- * Check if a path is ignored by the ignore files.
- * @param patterns - See the supported [glob patterns](https://github.com/sindresorhus/globby#globbing-patterns).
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -205,3 +205,42 @@ export function isGitIgnored(options?: GitignoreOptions): Promise<GlobbyFilterFu
 export function isGitIgnoredSync(options?: GitignoreOptions): GlobbyFilterFunction;
 
 export function convertPathToPattern(source: string): FastGlob.Pattern;
+
+/**
+ * Check if a path is ignored by the ignore files.
```
sindresorhus, line 210: Drop the leading `*` on each line, as with the existing doc comments.

sindresorhus:
Sorry, missed this.

index.d.ts
jiawei397: Modified it.

